PR #90, opened 2015-02-10
Enforce build with Visual Studio 2013

aveminus:
Environments with multiple Visual Studio installation tries to build with oldest available Visual studio.
So, if someone has Visual studio 2010 installed too, then it will throw error on c11 standard code (eg: override and range based loops).


kjk:
The change looks good but the PR includes more than just that.

aveminus:
I think I only changed the project files. Do you mean the merge commit ?

PR #163, opened 2015-04-11
fix issue with file encoding

CRVV:
Today I forked this project and compile it with Visual Studio 2013 Community 
But there are some compile errors
Then I wrote a [tool](http://github.com/CRVV/dcc) and try to fix them
The tool is not perfect so I use it to do most of work and manually fix the rest
This is the commit


kjk:
The code should compile as-is in VS 2013 Community. Are you using non-english OS? 

I will not merge this change. Most of the changes are not related to the code that is being compiled and from the diff I can't even tell what most of them change.

If you see a compilation error, then please describe what exactly are you compiling, post the error message and the minimal change that is required to fix that error on your computer.

C compilers were created before utf8 happened so converting the code to utf8 is probably not the right thing to do.

CRVV:
Yes, I am using Chinese OS.
The problem is simple. On Chinese OS, Visual Studio will decode non-unicode character as GB2312 and there are some decoding errors.
The minimal change that is required to fix that error on my computer is replacing those bytes with the Unicode substitution character. And I can not recognize them no longer.
Not only utf-8, the first C compiler were created before ISO-8859-1 and windows-1252 happened, too.
Finally, I do not care whether you will merge this change. I just think most people prefer unicode and I convert non-unicode to unicode. And then I can compile this project and I can add new feature to it.

PR #178, opened 2015-04-25
Fixed #61 - two-finger scrolling on touchpads should now work correctly

### LINKED ISSUE #61: Scrolling with Trackpad
Running SPDF 3.1.10094z on Win 8.1 x64 Pro/Surface Pro 3.

When I try to scroll using the two-finger swipe on the touchpad of the TypeCover keyboard, SPDF registers the scroll request (the scroll bar on the right of the window moves appropriately) but the view of the document does not change.

At the same time, using the scroll wheel of the mouse and the touch screen scrolls the document perfectly ok.

Philpax:
Two-finger scrolling gets handled as a SB_THUMBTRACK message; however, si.nPos and si.nTrackPos  are equivalent, which results in the canvas not being scrolled. This fix always scrolls the canvas if the message is SB_THUMBTRACK.

### DIFF
--- a/src/Canvas.cpp
+++ b/src/Canvas.cpp
@@ -81,7 +81,8 @@ static void OnVScroll(WindowInfo& win, WPARAM wParam)
     if (!IsContinuous(win.ctrl->GetDisplayMode()) && ZOOM_FIT_PAGE == win.ctrl->GetZoomVirtual())
         lineHeight = 1;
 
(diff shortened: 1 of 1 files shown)


kjk:
Thanks for the fix. Feel free to add yourself to AUTHORS file (names are in alphabetical order)

PR #205, opened 2015-05-22
Update Print.cpp

b4ufreak:
Added new -print-settings values:
color - print in color
monochrome - print in black and white
paper=<paper format name> - select paper size by paper format name

Added method to extract paper format name from size of page from the source file.  When printing a file from command line the proper paper size will be selected based on the format stored in the file.

### DIFF
--- a/src/Print.cpp
+++ b/src/Print.cpp
@@ -626,6 +626,52 @@ Exit:
     GlobalFree(pd.hDevMode);
 }
 
+static short GetPaper(BaseEngine *engine) {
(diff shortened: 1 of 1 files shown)


kjk:
Thanks. There are minor issues (using lstrcmpi instead of str::EqI and whitespace) which I'll fix.

If you wish, you can add yourself to AUTHORS file (https://github.com/sumatrapdfreader/sumatrapdf/blob/master/AUTHORS)

PR #244, opened 2015-07-18
new menu item: Duplicate Document

confusedsushi:
I often need quickly jump between two (or even more) pages in a document, for that it is very nice to have one document open multiple times.


kjk:
Thanks but menu items are reserved for functionality that is used very frequently and this is not such functionality.

PR #312, opened 2015-10-08
Do not display soft hyphens

### LINKED ISSUE #241: EPUB hyphenation
I find it really annoying, so I copy here issue no 2738 from the old google code tracker. The problem is that the program should not make the possible hyphenation marks visible (UTF8 character C2 AD) when UseFixedPageUI is set to false in the advanced options.
Example EPUB file: http://wolnelektury.pl/media/book/epub/sienkiewicz-sachem.epub

campbeb:
Most standards agree soft hyphens should not be displayed when not used
to break words at the end of a line.
Fixes #241


campbeb:
I understand if this is preferred to be implemented in a different manner and I am open to feedback.

kjk:
Thanks!

PR #313, opened 2015-10-09
Update Settings URL

campbeb:
Change URL in settings file to match website


kjk:
This actually needs to be versioned (i.e. settings3.1.html). I just haven't put the file on the server yet.

Settings might change between versions so we have a separate html for each version so that if you're in version 2.3, you're not seeing documentation for version 3.5.

settings.html is a copy of the latest released version to be linked from other docs.

Also, this file is auto-generated via gen_settingsstructs.py, as per comment at the top.

PR #316, opened 2015-10-10
Always let user use previous to get back to page 1

### LINKED ISSUE #315: Unable to use previous page to return to page 1 in epub with facing page view
Steps: 
 - Open an epub
 - Use facing page view (Menu / View / Facing)
 - Use Menu / Go To / Page and enter even numbered page
 - Use left arrow (or Page Up or menu command) to goto to previous page
 - Continue to beginning of epub
 - Can never get to page 1, only page 2

Go To First Page will take you back to page 1.

campbeb:
If showing double pages, user could not use previous to get to page 1,
only back to page 2.
Fixes #315

### DIFF
--- a/src/EbookController.cpp
+++ b/src/EbookController.cpp
@@ -562,9 +562,9 @@ bool EbookController::GoToPrevPage(bool toBottom)
 {
     UNUSED(toBottom);
     int dist = IsDoublePage() ? 2 : 1;
-    if (currPageNo - dist < 1)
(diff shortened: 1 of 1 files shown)


kjk:
Thanks. To keep things consistent between branches, I cherry-picked Simon's commit 0edae929510cde5f39d935e0907a610f0f0e37f2 from rel3.1working to master.

PR #375, opened 2015-12-02
Disable warning 4819 to prevent compiler error

shuhari:
It is not a bug, but a Visual C++ issue. Building the project on non-English OS would cause the following error:

warning C4819: The file contains a character that cannot be represented in the current code page (936). Save the file in Unicode format to prevent data loss 

I checked the source and found the non-standard characters are in comments, which may not cause any problem. So I add W4819 to the project's ignore list of warnings.

### DIFF
--- a/vs2015/enginedump.vcxproj
+++ b/vs2015/enginedump.vcxproj
@@ -142,7 +142,7 @@
       <PrecompiledHeader>NotUsing</PrecompiledHeader>
       <WarningLevel>Level4</WarningLevel>
       <TreatWarningAsError>true</TreatWarningAsError>
-      <DisableSpecificWarnings>4127;4324;4458;4800;%(DisableSpecificWarnings)</DisableSpecificWarnings>
+      <DisableSpecificWarnings>4127;4324;4458;4800;4819;%(DisableSpecificWarnings)</DisableSpecificWarnings>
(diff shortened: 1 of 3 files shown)


kjk:
The right place to make those changes is in https://github.com/sumatrapdfreader/sumatrapdf/blob/master/premake5.lua. Please read https://github.com/sumatrapdfreader/sumatrapdf/wiki/Premake5-build-system

Also, they should only be disabled for projects that actually fail to compile, not everywhere. We already disable those warnings for projects we know fail to compile.

shuhari:
Get the point. I should modify premake5.lua first.

I believe that you can compile successfully as long as the system culture is set to english, but it failed on my workstation (Windows 7 64 bit codepage Chinese 936) and Visual Studio 2015 SP1. As I see, there are 3 files with problem:

<code>mupdf/source/pdf/pdf-write.c  line#2075    >>  fprintf(opts->out, "%%\316\274\341\277\246\n\n");</code>

It seemed a little strange. Why the character code > 255?  I made some tests and found that VS2015 SP1 will cause warning about the line, while VS2013 not. I don't know how VS2015 (without SP)  will handle it.

<code>synctex_parser.c line#1058   >>   *  ?.123456789e123</code>
There is just a character which cannot show in my codepage.

<code>EngineDump.cpp line#307    >>  Out(UTF8_BOM);</code>
It seemed to be no problem, but VS2015 (SP1) also generate warning. I don't know why.

PR #431, opened 2016-02-09
add shift+scroll to scroll faster = half a page per tick

MateuszZarkowski:
I wanted faster scrolling for large documents. Very useful for quick navigation.
In other programs shift+scroll does that, it was free here, so I used it.
Cheers :)


kjk:
Thanks!

PR #487, opened 2016-04-04
Night mode

bi0ha2ard:
Added a second set of text and background colors that can be toggled with Ctrl+N

I don't know how your convention for the defines in resources.h is, so I just set it to a random number.


kjk:
Thanks for the patch.

I agree that customizing colors and an easy way of switching to other set (like night mode) is valuable, I'm not sold on doing it this way.

I'm very reluctant to add new menu items - we have too many of them already - and the guideline is that a new menu item would have to introduce significant feature that would be used by a lot of users.

I do have ideas on how more functionality could be added without overwhelming the UI for regular use, but there's nothing there yet for a foundation.

I'll keep this feature in mind but for now I'll not be merging this change.

PR #511, opened 2016-05-03
Enable DDE for plugin mode

StarGate01:
I cannot see why communication over DDE is disabled when SumatraPDF is launched in plugin mode (-plugin command line switch). I tested it, it works flawlessly. Please merge or explain.

### DIFF
--- a/src/SumatraPDF.cpp
+++ b/src/SumatraPDF.cpp
@@ -4256,8 +4256,8 @@ InitMouseWheelInfo:
             break;
 
         case WM_DDE_INITIATE:
-            if (gPluginMode)
-                break;
+            //if (gPluginMode)
+            //    break;
             return OnDDEInitiate(hwnd, wParam, lParam);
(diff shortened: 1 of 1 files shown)


kjk:
Plugin mode was created for embedding inside the browser as NPAPI plugin, so that you can open a PDF directly in the browser when clicking a link to a PDF file.

The whole code should be removed because we no longer support NPAPI plugin mode because browsers no longer support NPAPI plugins.

DDE is for integration with Latex editors and that use case doesn't apply when running as browser plugin.

StarGate01:
Ok, I do see your point. However, I am currently writing a Latex-plugin for Notepad++, and I am using SumatraPDF as an embedded PDF-reader. This is done by using the plugin mechanism. Now, to enable forward search etc., I send DDE messages to the child (plugin) window. 
It's perfectly fine with me if this is in your opinion a very unusual, retired or just not applicable use case and thus should not be merged, but it would come in quite handy.

PR #568, opened 2016-07-08
Make canvas scrolling DPI-aware

vanxining:
We should not use any absolute pixel values directly if HiDPI support is desired.


kjk:
Could you change it to use DpiScaleY() from Dpi.h for consistency with the rest of the code?

PR #643, opened 2016-10-19
Added Save Position button on menubar and in the context menu.

korypostma:
Sometimes I leave books open in Sumatra PDF and I need to go and do other things.  Before this change I had to close Sumatra PDF and then reopen it to save the position of where I was in the book.  With this change I just click the button.


kjk:
Sorry but this is not something I want to add. Menu items are precious resources, every addition makes the program a little bit worse so we reserve them for frequently used functionality. I wouldn't even know what "Save Position" means without reading an explanation first.

korypostma:
I never knew you were aiming for a minimal menubar.  What about the context menu or somehow making it save your reading position in case of crashes or computer restarts?

kjk:
It falls under "Simplicity of the user interface has a high priority." (http://www.sumatrapdfreader.org/free-pdf-reader.html).

We do save state when significant changes happen (like opening or closing a document, exiting the program). We don't save on every minor change (like scrolling) because that would be a lot of saving and crashes or restarts don't happen that often.

korypostma:
Since the view state is changed, iirc, why not have it check periodically if that state has changed and then save it.  It could check the cached view state and compare it every minute or few minutes and then only save if there was a change.  No need to do it upon every scroll.  Although I tested doing it with every scroll and it worked fine with no noticeable lag.  But I agree, that is excessive.  Checking it every now and then and then saving it if there is a change would fill both the need of keeping the UI simpler and also saving state in those rare cases.

kjk:
I'm not sure about that either. Again, it's about benefit vs. cost. 

Crashes don't happen that often.

If you test only with small history and a fast disk, saving of preferences is fast but since we remember every file opened, it accumulates and the size grows. And some people have sumatra (and settings file) on unpredictably slow network drives so if we keep saving periodically on main thread (as we do now) I'm pretty sure there will be people seeing the app freeze.

And if you move it to background thread, that adds complexity to ensure serialization and thread safety and making sure the save completes e.g. when user quits the app, so that we don't have half-written, corrupted preferences file. Currently prefs::Save() has to be called on main thread.

korypostma:
What about getting a windows lose focus message and then calling save when the window loses focus?  Maybe we should reconsider how we save this data to disk.  Anyhow, I'm just throwing ideas out there.

kjk:
That would be acceptable: on (WM_ACTIVATEAPP, false) schedule a delayed task by e.g. 1 min. Cancel it on (WM_ACTIVATEAPP, true). When 1 minute passes without activation, save prefs.

That should strike a balance of both ensuring crash resistance and not affecting performance.

PR #669, opened 2016-11-19
Lithuanian translation update

Romukas:
Added some missing strings.


kjk:
Thanks but please use http://www.apptranslator.org/app/SumatraPDF to submit translations. translations.txt is generated from apptranslator.org data.

PR #731, opened 2017-03-02
Theme Support

blake-mealey:
Hi!

I love the features of SumatraPDF but I am not a huge fan of the appearance. I just hacked together a simple theme system (using the existing preferences system) which you may want to use. If there is any refactoring you'd like me to do (I wasn't always sure where the correct place to put things was), I'd be happy to work on it and resubmit.

Here are some screenshots from the theme I made with my system:

sumatrapdf_2017-03-01_22-00-39
sumatrapdf_2017-03-01_22-01-11

One thing that doesn't quite work properly still is icon switching. In order for the other icon images to be loaded the build has to be cleaned and then rebuilt. I couldn't quite figure out how the image caching mechanism works and I had no idea how I could "fix" it without breaking functionality.

If you don't want this at all, that's fine of course, I'll just use it locally, although I think it would be great for others to get to customize the appearance a bit more!

Thanks


kjk:
Hi Blake, thanks for the PR. In general I like it. I admit Sumatra could use visual polish. Here's how I would like to approach this.

1. I like your icons better than the current one. At this point I don't want to expose it as a preference option (we actually already have a way to replace a toolbar by placing toolbar_11.bmp in the same directory as sumatra.exe). Let's just use new icons. If you could create a separate PR with just that change, that would be great.

2. For the sake of code simplicity (and long-term maintenance), I don't want to maintain multiple tab shapes. But I also have a preference for a square, simple shape. If you could prepare a separate PR that just changes the shape of the tab to square.

3. I'm happy to support themes but I don't want to expose them as preferences. Adding preferences  creates a pressure for maintaining future backwards compatibility and at this point we don't yet know if this is the right/full set of settings.

Also, I think a bigger positive impact would come from having 2 (or more) built-in themes that are easy to switch between than giving the user a way to create a theme from scratch.

In concrete terms, I would welcome the following PR:
* keep struct Theme
* define at least 2 themes ("light" and "dark" are a good start)
* add Settings/Theme menu with sub-items "Light" and "Dark" for switching between themes
More themes are also welcomed (could be e.g. based on Visual Studio Code themes).

For now I would be comfortable telling people that if they want to modify the look they have to edit the code and they could submit the theme for inclusion in Sumatra. That would keep the doors open to having full customizability in settings file in the future.

Implementation comment: currently in Sumatra we have code that tries to select colors of various things based on the Windows color theme (things like win->caption->bgColor in Tabs.cpp). That was a mistake because it complicates the code and sometimes the colors picked are really bad and there's no way for us to test all possible variations). Adding theming is a good opportunity to remove that code (basically remove support for UseSysColors setting and all the conditional code to support it).

In similar vein, there are some other things that are basically superseded by themes, like MainWindowBackground etc. There are probably more color-related settings that should be substituted but probably not all of them (the difficulty here is that we have 3 different uis (fixed page, comic book and ebook ui), which currently have different colors. Not sure if we should simplify and have theme colors apply to all of them, allow for a single theme to specify different colors for each ui or have Settings/Fixed Page Theme/*, and Settings/Ebook Theme/*  i.e. we have a set of themes but the 2 main UIs can have them chosen separately (comic book ui is so similar to fixed page ui that probably doesn't deserve it's own thing).

Comment on look of tabs: a common complaint is that it's hard to figure out which tab is currently selected. Colors in the screenshots are probably better than what we have currently but I think the best visual hint is when the background color of selected tab "bleeds" into the reset of the app, like in Visual Studio Code or Edge.

4. Finally, I'm sure there are other visual things that could be improved (like appearance of our notifications). I'm happy to merge more improvements.

blake-mealey:
Thanks for the feedback!

I have a few questions before implementing those requests:

1. The icon pack I used (https://material.io/icons/) looks much better with transparency, but obviously I had to modify them to work with the BMP format. I wasn't able to determine whether it is possible to switch to an image format which does support transparency. If you know that it is possible and could point me in the right direction, I'd be happy to implement that before submitting that PR.
2. I agree the square tabs are quite nice, however I think they would look a lot better if they were sharper. Currently, there is a distinct border surrounding the tabs which seems to be caused by some kind of anti-aliasing. I tried to change the render mode (https://github.com/sumatrapdfreader/sumatrapdf/blob/master/src/Tabs.cpp#L203) but that resulted in nothing being rendered at all. If you could point me in the right direction to fix this I'll do so, otherwise I can just submit a PR as is.
3. Where should I put the various theme structs (light, dark, etc)? Should I keep the current theme stored in gGlobalPrefs->theme? It seems like that would mean it is still part of the preferences file, but I also do not know how else to save current preferences. I suppose there could be a theme string in gGlobalPrefs which saves the current theme, but then the actual theme struct should be easily accessible from everywhere else, and I do not know where I would store it.
4. I was trying to figure out if there is a way to change the colours (background and text) of the menus. For a proper dark theme, we would need to change the menus as well, although I understand this may be difficult since you are using the Windows APIs.

kjk:
Re 1: I've added comments to https://github.com/sumatrapdfreader/sumatrapdf/issues/734

kjk:
Re 2: https://github.com/sumatrapdfreader/sumatrapdf/issues/735

kjk:
Re 4: it's possible, see e.g. https://code.msdn.microsoft.com/windowsapps/CppWindowsOwnerDrawnMenu-a9888b0c or google for "owner drawn menu" or "custom drawn menu".

VictorVG:
Initially, the icons seem unusual, but a closer look - comfortable and like the eyes do not get tired. A background color - it I am always under her mood. In general, this option seems best. :+1:

kjk:
Closing as it's been split into several independent issues.

Issue #727, opened 2017-02-23
Keyboard shortcut for Favorites


Arklytte:
I make/utilize a significant number of PDF's daily that dont have Bookmarks.  Being able to quick add favorites would be a HUGE timesaver.

FYI, I'm a HUGE fan of your software, and have been using it for a couple of years.  I use it exclusively at work and at home and find it to be (mostly) far more user friendly than it's competition.  However, the lack of user customizability for keyboard shortcuts is very frustrating.

kjk:
"Alt - a" followed by Enter works as a shortcut.

Closing as duplicate of #265

PR #737, opened 2017-03-04
Resolving issue #727: Keyboard shortcut for Favorites

H3xag0n:
Added hotkey Ctrl+B for adding current page to Favorites (with the dialog interactively).
Unfortunately, I don't know how to use scripts\trans_gen,py to update translations. It's neccessary for showing up the hotkey title in main menu -> Favorites -> Add to Favorites


kjk:
Thanks, but while Ctl-B seems like a reasonable shortcut, our philosophy is to match what other similar software does but I don't have the data on that, so I don't want to introduce a shortcut without doing such legwork.

In other works, why Ctrl-B and not some other shortcut? What does Adobe Reader use? What does FoxIt use? What do other document readers use? What do other programs use?

Issue #736, opened 2017-03-03
Implement theming for changing look of Sumatra


kjk:
@blake-mealey continuing from https://github.com/sumatrapdfreader/sumatrapdf/pull/731

> Where should I put the various theme structs (light, dark, etc)? Should I keep the current theme stored in gGlobalPrefs->theme? It seems like that would mean it is still part of the preferences file, but I also do not know how else to save current preferences. I suppose there could be a theme string in gGlobalPrefs which saves the current theme, but then the actual theme struct should be easily accessible from everywhere else, and I do not know where I would store it.

For now I would just make them as globals in `Menu.cpp`, since they are switchable from menu.
```
struct Theme { };

Theme g_themeLight = { ... };
Theme g_themeDark = { ... };

#define THEME_COUNT 2

Theme *g_themes[THEME_COUNT] = {
    g_themeLight,
    g_themeRight
};
```

If it becomes big, theme definitions could be moved to `Theme.cpp`.

We would need to add one setting "themeName" to remember selected theme, similar to gGlobalPrefs->uiLanguage.

blake-mealey:
So I am having a few difficulties getting this to work simply from an organizational standpoint. I have the structs declared very similarly to how you described doing it above in `Menu.cpp`, I created the Settings > Theme > [...] menu, and I added the `themeName` property to `GlobalPrefs`. But now I need a way to get the current theme which is accessible from everywhere that `gGlobalPrefs` is. I was thinking of doing something like this in `GlobalPrefs.h`:

```
inline Theme* GetTheme() {
    for (int i = 0; i < THEME_COUNT; i++) {
        if (!strcmp(g_themes[i]->name, gGlobalPrefs->themeName)) {
            return g_themes[i];
        }
    }
}
inline void SetTheme(Theme *newTheme) {
    gGlobalPrefs->themeName = newTheme->name;
}
}
```

But, Theme, THEME_COUNT, g_themes, etc. are not accessible from `GlobalPrefs.h`. Is there a better place I can put them?

Also, if I update `gGlobalPrefs->themeName` during runtime, will it save the preferences file with that change, or do I need to call another method to do that?

kjk:
Generically speaking, you can always export the things you need in one place from another place.

In this particular place I would have `Theme *GetThemeByName(char* name)` in `Menu.cpp`. 
[...]

Issue #735, opened 2017-03-03
Improve the look of tabs


kjk:
@blake-mealey continuing from https://github.com/sumatrapdfreader/sumatrapdf/pull/731

> I agree the square tabs are quite nice, however I think they would look a lot better if they were sharper. Currently, there is a distinct border surrounding the tabs which seems to be caused by some kind of anti-aliasing. I tried to change the render mode (https://github.com/sumatrapdfreader/sumatrapdf/blob/master/src/Tabs.cpp#L203) but that resulted in nothing being rendered at all. If you could point me in the right direction to fix this I'll do so, otherwise I can just submit a PR as is.

I didn't write code in Tabs.cpp and I'm not GDI+ expert either. I believe GDI+ was used to get anti-aliased rounded corners of the tab, which is not possible with GDI.

If it's easier to do things in GDI, you can replace GDI+ code with GDI.

kjk:
Fixed by https://github.com/sumatrapdfreader/sumatrapdf/pull/740

PR #740, opened 2017-03-07
Implement theming

blake-mealey:
To take care of: https://github.com/sumatrapdfreader/sumatrapdf/issues/736

I used the VS UI to modify resource.h. It seems to have created quite a mess in terms of created and modified files. Hopefully that's okay.

I realized as I went I could expand on the `Theme` struct quite a bit (made it more modular, added some fields) and ended up creating `Theme.h` and `Theme.cpp` because the theme stuff was quite large. I removed lots of old code in `Tabs.cpp` which formed the only existing theme and replaced it with a more powerful and general tab theming system.

I removed support for `UseSysColors` and `MainWindowBackground`. I removed `FixedPageUI/EBookUI.TextColor/BackgroundColor` and now use `currentTheme->document.textColor/backgroundColor`. I wasn't sure how to handle the changes in `AppPrefs.cpp` since the comments there already mentioned backwards compatibility and now I am completely removing those properties.

I am not sure if it was the best idea to replace `FixedPageUI/EBookUI.TextColor/BackgroundColor` as I did, simply because it would be nice to get default document colours even with a dark theme on. Maybe I could add another menu item under themes which enables or disables this.


kjk:
Please undo the changes to resources file. It doesn't look like anything has actually changed but a bunch of the changes that VS did are incorrect (like replacing macros with their values).

It's unfortunate that we can't use VS to edit resource file but it is what it is. I put as little as possible there to make editing with a text editor bearable.

src/Menu.cpp
```diff
@@ -701,8 +703,18 @@ HMENU BuildMenu(WindowInfo *win)
         RebuildFavMenu(win, m);
         AppendMenu(mainMenu, MF_POPUP | MF_STRING, (UINT_PTR)m, _TR("F&avorites"));
     }
+
+    // Build the themes sub-menu of the settings menu
     m = BuildMenuFromMenuDef(menuDefSettings, dimof(menuDefSettings), CreateMenu(), filter);
+    MenuDef menuDefTheme[THEME_COUNT];
+    static_assert(IDM_CHANGE_THEME_LAST - IDM_CHANGE_THEME_FIRST + 1 >= THEME_COUNT, "Too many themes. Either remove some or update IDM_CHANGE_THEME_LAST");
+    for (UINT i = 0; i < THEME_COUNT; i++) {
+        menuDefTheme[i] = { _TRN(GetThemeByIndex(i)->name), IDM_CHANGE_THEME_FIRST + i, 0 };
```
kjk: The way translation system works is:
* _TRN() marks a literal string that needs to be translated. Here you need to call `trans::GetTranslation(GetThemeByIndex....)`

src/Theme.cpp
```diff
@@ -0,0 +1,301 @@
...
+#include "SettingsStructs.h"
+#include "GlobalPrefs.h"
+
+#define COL_BLACK 0x000000
+#define COL_WHITE 0xFFFFFF
+
+// Themes
+Theme g_themeLight = {
+    // Theme Name
+    "Light",
```
kjk: As noted above, this needs to be marked with _TRN("Light").

src/resource.h
```diff
@@ -1,208 +1,209 @@
-//{{NO_DEPENDENCIES}}
```
kjk: This whole file is marked as changed. Please edit in text editor so that only real changes show up in the diff.

vs2015/SumatraPDF.vcxproj.filters
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
```
kjk: Please undo changes here. Those files are auto-generated from `premake5.lua` and will be over-written on next generation. You need to add Theme.cpp/Theme.h to list of files in premake5.files.lua and re-generated files as described at the top of premake5.lua.

vs2015/sumatra.vcxproj.filters
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
kjk: Also should be a result of premake5 generation.

src/Menu.cpp
blake-mealey: Since `MenuDef` expects `char *`, and `trans::GetTranslation` returns `WCHAR *`, what is the correct way to convert `WCHAR *` to `char *`?

blake-mealey: Never-mind, I found `str::conv::ToUtf8`

src/AppPrefs.cpp
```diff
@@ -240,10 +240,10 @@ void UpdateGlobalPrefs(const CommandLineInfo& i) {
                    str::EqI(i.globalPrefArgs.At(n), L"-bg-color")) {
             // -bgcolor is for backwards compat (was used pre-1.3)
             // -bg-color is for consistency
-            ParseColor(&gGlobalPrefs->mainWindowBackground, i.globalPrefArgs.At(++n));
+            //ParseColor(&gGlobalPrefs->mainWindowBackground, i.globalPrefArgs.At(++n));
```
blake-mealey: This seems rather messy, and may be part of a larger backwards compatibility problem.

src/SumatraPDF.cpp
```diff
@@ -2001,30 +2002,18 @@ static void RerenderEverything()
...
 {
-    if (gGlobalPrefs->useSysColors) {
-        text = GetSysColor(COLOR_WINDOWTEXT);
-        bg = GetSysColor(COLOR_WINDOW);
-    }
-    else {
-        text = gGlobalPrefs->fixedPageUI.textColor;
-        bg = gGlobalPrefs->fixedPageUI.backgroundColor;
-    }
+
```
blake-mealey: I think this (and GetEbookUiColors) should have a toggle to use either the current theme's colours or default (black and white). This toggle should be global though, not individual for ebookUI and fixedPageUI.

src/Menu.cpp
```diff
@@ -701,8 +703,18 @@ HMENU BuildMenu(WindowInfo *win)
         RebuildFavMenu(win, m);
         AppendMenu(mainMenu, MF_POPUP | MF_STRING, (UINT_PTR)m, _TR("F&avorites"));
     }
+
+    // Build the themes sub-menu of the settings menu
     m = BuildMenuFromMenuDef(menuDefSettings, dimof(menuDefSettings), CreateMenu(), filter);
+    MenuDef menuDefTheme[THEME_COUNT];
+    static_assert(IDM_CHANGE_THEME_LAST - IDM_CHANGE_THEME_FIRST + 1 >= THEME_COUNT, "Too many themes. Either remove some or update IDM_CHANGE_THEME_LAST");
+    for (UINT i = 0; i < THEME_COUNT; i++) {
+        menuDefTheme[i] = { str::conv::ToUtf8(_TR(GetThemeByIndex(i)->name)) , IDM_CHANGE_THEME_FIRST + i, 0 };
```
kjk: Sorry, I misled you. It should just be:
```
menuDefTheme[i] = { GetThemeByIndex(i)->name , IDM_CHANGE_THEME_FIRST + i, 0 };
```

`BuildMenuFromMenuDef()` calls `GetTranslation()` (unless `MF_NO_TRANSLATE` flag is set).

src/Theme.cpp
```diff
@@ -0,0 +1,315 @@
...
+    for (int i = 0; i < THEME_COUNT; i++) {
+        if (str::Eq(g_themes[i]->name, name)) {
+            return g_themes[i];
+        }
+    }
+    return NULL;
+}
+
+Theme *GetThemeByIndex(int index) {
+    return g_themes[index];
```
kjk: In places like this it sometimes makes useful to add: `CrashIf(index < 0 || index >= THEME_COUNT)`, given that C lacks array bounds checking.

CrashIf() is only compiled in debug and pre-release builds and would trigger a crash.

src/SumatraPDF.cpp
kjk: How do you see this toggle, as a menu item?

I don't have an opinion here. I would have to use the app with theme support for a bit to get a feel for how it works.

Also we can do a pre-release with this new code and ask people for feedback on the forum.

kjk:
It's not required to land this change, just noting that the way notifications look (e.g. when you do a search in pdf file) should probably also conform to the theme. The code is in Notifications.cpp.

src/SumatraPDF.cpp
blake-mealey: I was imagining a menu item like `Settings > Theme > Theme Document Colors`. A pre-release seems like a good idea. It will also give the opportunity to modify the current theme colours, especially light, which I just left exactly as it was before and could be improved on (like you were saying it is hard to tell which tab is the active tab in that theme).

blake-mealey:
I just made those fixes and added notification theme support. I do not think I have seen all possible notifications so they may not all look good. Here is a screenshot with the darker theme:

kjk:
Looks good, thanks. I'll merge it later today.

kjk:
Actually, I've made changes to Notifications.cpp on the master  and your changes seems to be based on earlier code so it won't merge cleanly.

Could you rebase against the master and fix merge conflicts. It might actually easier to re-do the changes  in Notifications.cpp since the changes in master were pretty extensive (although mostly cosmetic).

blake-mealey:
Hmm. I am not 100% confident I could do that properly with my current Git skills. Maybe I could go back a few commits on this branch, and you can merge it. Then I'll create a new PR with notification theming changes.

A couple other things. I just realized I made the tabs square in this PR and not in a separate one for issue https://github.com/sumatrapdfreader/sumatrapdf/issues/735, and I just fixed that AA problem that I was talking about for that issue. Should I commit that change here as well? Or a separate PR?

kjk:
Ok, then don't worry about it. I'll grab this change, make fixups and commit.

No need to make separate PRs, we'll just close those issues with this PR.

kjk:
Thanks, I've merged the changes after some git massaging.

kjk:
I've released a pre-release build with those changes (http://www.sumatrapdfreader.org/prerelease.html)

PR #741, opened 2017-03-07
Remove highlight by double 'h'

zklhp:
Select the highlighted part, press 'h', the highlight removed.

The logic turns to first check if there is a corresponding highlight, if there is, delete; if not, add.

It is a bit dirty, and only tested with PDF files.

### DIFF
--- a/src/SumatraPDF.cpp
+++ b/src/SumatraPDF.cpp
@@ -3547,7 +3547,15 @@ static void FrameOnChar(WindowInfo& win, WPARAM key, LPARAM info=0)
             if (!win.AsFixed()->userAnnots)
                 win.AsFixed()->userAnnots = new Vec<PageAnnotation>();
             for (SelectionOnPage& sel : *win.currentTab->selectionOnPage) {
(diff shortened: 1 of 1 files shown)


kjk:
Thanks!

joshualyl:
Hello zklhp. 
Thanks for your reply and adding the feature of remove of highlighting.
There are some problem in using this feature. While I select an area by "ctrl + left click and drag", if I highlight this area, it was almost impossible to select it again, so it can't be remove. In addition, when I mark "abc", and select "ab" press 'h', it will highlight "ab" again.
As some PDF files include pictures, and mark an area is very useful, it would be better if the highlights could be selected.

zklhp:
Hi joshualyl,
Ops. I have not tested with the Ctrl selection case. In fact, I didn’t know it exists. So, unfortunately, it is out of my ability.

For the second case, yes, only if you select the “abc” again can you remove the highlight.

In my understanding, the code of highlight and selection are different, so linking them needs some work. I think an independent UI is needed like what Adobe’s software do.

Or we can have a short key, such as “Ctrl+Alt”, when holds it and click the area that has highlight, it will be removed. Either way, a design of the feature is needed.

P.S. There are actually some designs, such as

http://nod5.dcmembers.com/sumatra_highlight_helper.html

Sorry for the inconvenience.

joshualyl:
Thanks !

PR #751, opened 2017-03-16
Compilation fixes

hadess:
Hopefully the DOS line endings didn't get mashed up. Let me know if this directory (`ext/unarr`) is the upstream or if there is another one.


kjk:
You should probably submit those fixes to https://github.com/zeniko/unarr.

hadess:
> You should probably submit those fixes to https://github.com/zeniko/unarr.

Does that mean that the read-only mirror of this directory is actually its upstream? I don't understand.

kjk:
Well... zeniko is the author of this code. He also used to work on Sumatra and at that time that was true. Zeniko doesn't work on Sumatra anymore so I don't know how he wants to manage unarr development but he is the de-facto maintainer of unarr code by the virtue of being its author so his repo is de-facto upstream of this code.

Issue #657, opened 2016-11-09
Toolbar: add "BACK|Forward" navigation button(s)


VitRom:
At present the only ways to navigate back/forward after a jump from TOC or anything like this are
1) keyboard hotkeys [Alt]+[Left/Right] and
2) Menu/Navigation/Back (/Forward) commands.
IMHO this type of navigation is a more frequent use case when reading a complex documents with a TOC and a lot of internal links. But the only navigation related buttons on toolbar are back/forward **_search_** and prev/next **_page_** that are useless in this case and forces user to use abovementioned ways

$subj

PS. And the context menu, currently with a "sel all" and "copy" commands only, is a right place for this navigation function too.

SumatraPeter:
https://github.com/sumatrapdfreader/sumatrapdf/wiki/Toolbar-design

VitRom:
Er... I'm not about bells and whistles and not about blowing things up with a glows, spikes, animations and misc UFO-like floating crap. Just about "a quick access to frequently used functionality" and what exact this "frequently" is.

I and most of my colleagues reads a 99% of PDFs in the "Continious scroll" mode with a "Fit to page width" zoom. Most of times it's a different kinds of manuals and userguides viewed in a normal non-maximized windows placed aside to another windows we're working with. And a much faster way to go to the next/previous page is a brief scroll with a mouse wheel. So these two "Prev Pg/Next Pg" buttons are useless.

At the same time these abovementioned manuals contains a misc links, i.e. TOC, inter-references, footnotes. And after a several jumps through these links reader have to recall a page number where he has started his digging and either type it in the "Page:" filed either just scroll. This case these two buttons mentioned above are useless too.

And now pls review a described above assuming these two buttons calls a function I've asked for, "Prev/Back". Still no overbloat but much handy, isn't it?

SumatraPeter:
Oh, don't get me wrong, I do think the toolbar needs an overhaul or ideally be user configurable. However I just provided the wiki link above to inform you about how the developers look at all such toolbar redesign requests.

shaiay:
I made a very simple change to add this: https://github.com/shaiay/sumatrapdf/commit/413bf0ba07c6196127185451a35b12e247672585
However, currently it uses the same icons as prev & next page, so a bit confusing. Any help with nicer icons would be appreciated. Once I have better icons I will create a pull request

[...]

Issue #734, opened 2017-03-03
Improve toolbar icons


kjk:
@blake-mealey continuing discussion from https://github.com/sumatrapdfreader/sumatrapdf/pull/731

> The icon pack I used (https://material.io/icons/) looks much better with transparency, but obviously I had to modify them to work with the BMP format. I wasn't able to determine whether it is possible to switch to an image format which does support transparency. If you know that it is possible and could point me in the right direction, I'd be happy to implement that before submitting that PR.

Currently we use standard toolbar control. It only supports .bmp files but it does support transparency (pixels RGB(0xFF, 0, 0xFF) in the .bmp image are transparent in the toolbar).

My wishlist is to stop using windows toolbar control and instead paint the icons in code, preferably using a textual vector format, which would give us transparency and good looking dpi scaling for free.

We already have code in SvgPath.cpp to parse a subset of SVG path expression and draw it using GDI+, so it should be easy to extract relevant parts of the icon from .svg file and use in Sumatra.

blake-mealey:
Would moving away from the windows toolbar control also involve this stuff? (https://code.msdn.microsoft.com/windowsapps/CppWindowsOwnerDrawnMenu-a9888b0c, https://msdn.microsoft.com/en-us/library/windows/desktop/ms647558(v=vs.85).aspx#_win32_Creating_Owner_Drawn_Menu_Items)

I don't have any prior experience with these libraries but I'd love implement a custom toolbar that would allow for more customization and better looking icons.

kjk:
As background, with CreateWindow* you can either create "raw" windows, where you have to provide your WndProc (a function that processes windows messages sent to a window). Then you draw into that window on WM_PAINT message.

Or you can create one of the custom controls, like menu control or toolbar control. Those are also windows but the os paints them, provides most of the low-level handling and eposes a nicer, high-level interface (where "nicer" being relative; it's less work that doing everything from scratch but the API to custom controls is not how we would write code today).

Some custom controls allow for certain levels of customization. The above articles describe how to modify the look of standard menu control (as complicated as it looks, it's much less work than re-implementing menus from scratch).

For toolbar control, I don't think it allows for enough paint customization so it has to be done from scratch. There are 2 possibilities.

1. Write HWND-based control, similar in spirit and functionality to standard toolbar control. See LabelWithCloseWnd.[h|cpp] for a template of such style and an implementation of a simple control in this style.

2. HWND-less implementation.

[...]

PR #781, opened 2017-04-15
Use toolbar left/right arrows for prev/next or back/forward (according to options)

shaiay:
Solves Issue #657

### DIFF
--- a/src/SettingsStructs.h
+++ b/src/SettingsStructs.h
@@ -365,9 +365,11 @@ struct GlobalPrefs {
     // if true, we show a list of frequently read documents when no
     // document is loaded
     bool showStartPage;
-    // if true, documents are opened in tabs instead of new windows
-    bool useTabs;
-    // information about opened files (in most recently used order)
+	// if true, documents are opened in tabs instead of new windows
+	bool useTabs;
+	// if true, tollbat arrows do BAck/FOrward instead of Prev/Next
+	bool toolbarBackForward;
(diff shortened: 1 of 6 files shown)


kjk:
Thanks for the PR but I don't want to do it this way.

Adding one more setting to change how a program works is not a viable long-term strategy. Every additional setting makes the program a little bit worse and all other settings harder to find.

This is not to say I'm opposed to exposing back/forward in toolbar or some other easily accessible way but doing it right is exponentially harder than adding another setting.

shaiay:
Do you have any thoughts on how to do it right?

kjk:
Yes, but it's rather involved.

First, the toolbar itself needs to be more flexible. Currently the icons are in a bitmap. We need to be able to use vector icons (see https://github.com/sumatrapdfreader/sumatrapdf/issues/734) so that we can easily add more icons for more operations.

Second, make the toolbar configurable in settings file i.e. allow people to define a custom order by specifying order of icon/command associated with the icon.

shaiay:
I'm afraid this is too big a project for me right now. 
My fork is always available if someone needs this very specific functionality and can't wait until it is implemented right.

shaiay:
There seems to be a standard way to make a configurable toolbar:
https://msdn.microsoft.com/en-us/library/windows/desktop/hh298385(v=vs.85).aspx
This does not however move us away from bitmap icons

PR #793, opened 2017-04-30
Redo how application colors are defined

kjk:


VictorVG:
Christoph! Big thank you! The experiment with the coloring of tabbed headers on the one hand is interesting, and on the other hand, there was a feeling of somehow internal internal incompleteness, and as a result, only the problems and discontent of users were left of it. Habitual interface let someone do not like colors, but at least there are no problems with it. :up:

PR #819, opened 2017-06-01
Added rotation options to -print-settings command line flag

### LINKED ISSUE #475: printing rotation
I'm using SumatraPDF to print pdf by command line to printer Zebra LP 2824 Plus.
When the report is printed, it is rotated 90 degrees.
Is there a way to avoid the rotation?

greasypablo:
new -print-settings options are [autorotate|portrait|landscape]
autorotate is the current functionality
portrait forces portrait mode (x shorter than y)
landscape forces landscape mode (y shorter than x)

This should be useful with label printers and the like.

Hopefully resolves #475


kjk:
Looks reasonable but please do 2 cosmetic changes:
* squash the checkins into 1 (e.g.  with `git rebase --interactive`)
* format with `scripts\clang-format.ps1`

PR #820, opened 2017-06-02
Added rotation options to -print-settings command line flag (#1)

### LINKED ISSUE #475: printing rotation

greasypablo:
new -print-settings options are [autorotate|portrait|landscape]
autorotate is the current functionality
portrait forces portait mode (x shorter than y)
landscape forces landscape mode (y shorter than x)

This should be useful with label printers and the like

Hopefully resolves #475


greasypablo:
I've done my best to tidy this up but my spectacular unfamiliarity with Git, Powershell etc is not helping.

Hopefully this is good enough.

kjk:
Looks good, thanks for the patch!

PR #826, opened 2017-06-09
Fix search across pages

pcrama:
If you look at the last commit message, you will see that the code is not 100% ready to be merged, but I already wanted to see if you were interested in a patch to search for phrases (several words separated by space) across page boundaries (tested for PDF only).

Note that based on the testing infrastructure I added to this, I would also be interested in working on

- Trying to see if there is a visible performance increase in search by not looking for the anchor twice (currently, a page with a possible match is discovered by looking for the first word of the search text (=the anchor), then at those places, looking for the whole phrase. The first characters corresponding to the anchor could be skipped, I think)

- [this is more ambitious and I'm not sure I can pull it off] looking for an open source regular expression library that could be integrated

What are your thoughts?


pcrama:
The AppVeyor build fails with

```
fatal: reference is not a tree: 418af7d4a5e08e01fc7e49916c7a5f1aacc39c30
```

I'm sorry but I have no idea what I can do about it.  As far as I understand the error message, AppVeyor has trouble checking out the source, but I don't think I did anything strange with the repository that would warrant that.

kjk:
Don't worry about AppVeyor, it does that on every pull request and I don't know why either.

The functionality sounds useful.

Some thoughts:
* I don't plan on adding regular expression search. Even I am not a proficient regex user and Sumatra's target users are way less sophisticated in that regard than programmers
* making things faster is always good
* I'll leave specific comments on the code. Sumatra's code has its own mostly unspoken rules and I want to preserve that
* as a heads up: when the code is ready to merge, I'll ask you to squash multiple commits into one

src/TestTextSearch.cpp
```diff
@@ -0,0 +1,107 @@
+#include <iostream>
+#include <fstream>
+
+#include "BaseUtil.h"
+// layout controllers
```
kjk: this is a copy-paste. Granted the comment is misleading in the original place as well.

src/TestTextSearch.cpp
```diff
@@ -0,0 +1,107 @@
...
+    // an HtmlWindow and thus outside the reach of the main UI)
+    virtual void FocusFrame(bool always) { always; };
+    // tell the UI to let the user save the provided data to a file
+    virtual void SaveDownload(const WCHAR *url, const unsigned char *data, size_t len) { url; data; len; };
+    // EbookController //
+    virtual void HandleLayoutedPages(EbookController *ctrl, EbookFormattingData *data) { ctrl; data; };
+    virtual void RequestDelayedLayout(int delay) { delay; };
+};
+
+void _lossyWcharOut(std::ofstream &lf, const WCHAR *s)
```
kjk: We don't do `_foo`. It should be `lossyWcharOut`. If the intent is to indicate as private to the file, use `static`.

kjk: We don't use `std::ofstream` in the code base currently and I want to keep it that way. The `lf` should be `std::string`.

src/TestTextSearch.cpp
```diff
@@ -0,0 +1,107 @@
...
+{
+    for (const WCHAR *p = s; *p; ++p) {
+        lf << (((*p >= 32) && (*p <= 127)) ? ((char)(*p & 0x7f)) : '?');
```
kjk: I'm tempted to ban ? : because inevitably it leads to inscrutable code. This should be:
WCHAR c = *p;
if (c < 32 || c > 127) {
  c = '?';
}

src/TestTextSearch.cpp
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+void SearchTestWithDir(WCHAR *searchFile, WCHAR *searchTerm, std::ofstream &logFile, TextSearchDirection direction)
```
kjk: again, no `std::ofstream`

src/TestTextSearch.cpp
```diff
@@ -0,0 +1,107 @@
...
+            logFile << "findCount=" << findCount << "\n";
+            for (int i = 0; i < tsel->len; ++i) {
+                RectI rect = tsel->rects[i];
+                logFile << " pages[" << i << "]=" << tsel->pages[i]
+                        << " rects[i]=(" << rect.x << ", " << rect.y << ", "
+                        << rect.dx << ", " << rect.dy << ")\n";
+            }
+        } while ((tsel = tsrch->FindNext()) != nullptr);
+    }
+    double dur = t1.GetTimeInMs();
```
kjk: the if/do loop could be collapsed into a single for:
```
for (auto tsel = tsrch->FindFirst((FIND_FORWARD == direction)?1:engine->PageCount(), searchTerm); tsel != nullptr; tsel = tsrch->FindNext())) {
  ...
}

File: src/TestTextSearch.h [hunk 6]
```diff
@@ -0,0 +1,2 @@
+#pragma once
```
kjk: We don't use `#pragma once`.

src/TextSearch.cpp
```diff
@@ -116,23 +116,35 @@ void TextSearch::SetLastResult(TextSelection *sel)
...
 // try to match "findText" from "start" with whitespace tolerance
 // (ignore all whitespace except after alphanumeric characters)
-int TextSearch::MatchLen(const WCHAR *start) const
+TextSearch::PageAndOffset TextSearch::MatchEnd(const WCHAR *start) const
 {
     const WCHAR *match = findText, *end = start;
+    const PageAndOffset invalid = { -1, -1 };
+    int currentPage = findPage;
+    const WCHAR *currentPageText = pageText;
+#define _advance_page \
```
kjk: Inline the use of `_advance_page` and remove it.

src/TextSearch.cpp
```diff
@@ -116,23 +116,35 @@ void TextSearch::SetLastResult(TextSelection *sel)
 
 // try to match "findText" from "start" with whitespace tolerance
 // (ignore all whitespace except after alphanumeric characters)
-int TextSearch::MatchLen(const WCHAR *start) const
+TextSearch::PageAndOffset TextSearch::MatchEnd(const WCHAR *start) const
 {
     const WCHAR *match = findText, *end = start;
+    const PageAndOffset invalid = { -1, -1 };
```
kjk: `invalid` => notFound

src/TextSearch.h
```diff
@@ -30,6 +30,11 @@ class TextSearch : public TextSelection
     int GetCurrentPageNo() const { return findPage; }
 
 protected:
+    // Lightweight container for page and offset within the page to use as return value of MatchEnd
+    typedef struct _PageAndOffset {
```
kjk: No need for typedef in C++. Should be `struct PageAndOffset { ... }`

vs2017/SumatraPDF.vcxproj
```diff
@@ -335,6 +335,7 @@
     <ClCompile Include="..\src\SumatraPDF.cpp" />
     <ClCompile Include="..\src\SumatraStartup.cpp" />
     <ClCompile Include="..\src\Tests.cpp" />
+    <ClCompile Include="..\src\TestTextSearch.cpp" />
```
kjk: You seem to have manually updated the files. The right way is to change `premake5.lua` and/or `premake5.files.lua` and re-run premake to re-generate project files. See https://www.sumatrapdfreader.org/docs/Our-build-system-3114db56b1a54b9bb026da6b92d36a80.html

pcrama:
Thank you for the review.

> I don't plan on adding regular expression search...

Pity...
 
> making things faster is always good

Unfortunately, it doesn't give a measurable speedup and makes the code more complex, so I had to drop that idea.

> as a heads up: when the code is ready to merge, I'll ask you to squash multiple commits into one

I looked at your comments and thought it was easier to start from a fresh branch and copy things over.  The result should be more or less the same.  Is that OK?

> I don't want to add (possibly large) .pdf files to repository. My plan is to put them into sha1-addressable, publicly accessible location

Something like `git lfs`

> were .log files meant to be checked in?

Yes, I wanted to use them as regression cases to make sure I still found at least the same matches as the previous code.  I can maybe encode the same information inside the TestTextSearch.cpp as assertions, which would alleviate the need to diff the log files after the test run to check nothing changed.

Overall, I think your comments on the code are clear to me, so I will start on the changes.

Best regards,
`-- ` 
Philippe Crama

kjk:
> I looked at your comments and thought it was easier to start from a fresh branch and copy things over. The result should be more or less the same. Is that OK?

Yes, either way is fine.

FYI: to squash multiple commits into one: `git rebase --interactive`.

> Something like git lfs

Git lfs seems to be an extension that needs to be installed which is a burden for everyone who works with the code. I prefer if it's uploaded to s3/google storage/dropbox or what not.

If I have access to files, I can upload them to my backblaze-based system.

pcrama:
Hello,

I pushed a new branch: https://github.com/pcrama/sumatrapdf/tree/rework-fix-search-across-pages
That branch should address your concerns, I hope.

How do I get it into this pull request?

1. Cancel this pull request and create a new one
2. Force-push the new branch to pcrama:fix-search-across-pages
3. Something else (what?)

Best regards,
Philippe Crama

kjk:
The simplest would be: close this PR and create a new one

pcrama:
Closing this PR to reopen it with the new branch... See #844

PR #857, opened 2017-08-05
Change Frequently List to Auto Rows and Cols

SalmanZivari:
Before this change, SamutraPDF would showed only 2 rows, but now it can show maximum rows that size of window allows. another change was size of thumbnail! now it is half of what was before.


kjk:
This shows up as 3 new files at top-level. Probably meant to be changed files inside src/ directory.

SalmanZivari:
That's right. it must be in src directory

SalmanZivari:
should i upload again?!

SalmanZivari:
I moved files to src Directory. sorry for mistake

kjk:
It's better but it seems to be based on older version of the code which means that it undoes some recent changes.

src/SumatraStartup.cpp
```diff
@@ -796,7 +796,7 @@ int APIENTRY WinMain(_In_ HINSTANCE hInstance, _In_opt_ HINSTANCE hPrevInstance,
 
     retCode = RunMessageLoop();
     SafeCloseHandle(&hMutex);
-    CleanUpThumbnailCache(gFileHistory);
```
kjk: Why commented-out?

SalmanZivari:
I think now all things is correct!

src/SumatraStartup.cpp
SalmanZivari: because now the FILE_HISTORY_MAX_FREQUENT is not needed for displaying thumbnail on window!!! but CleanUp Method cleane by FILE_HISTORY_MAX_FREQUENT. this method must be rewrite with another algorithm if we want to clean up thumbnails.

SalmanZivari: in fact if we opened 40 file - and now for example can be showed on windows - but FILE_HISTORY_MAX_FREQUENT was 10 ... well, next time the 20 thumbnail doesn't exist!!!!! (we have 40 file in frequently list!! and last 20 file have not thumbnail!!)

kjk: Well, that seems to be needed to make the whole thing work?

SalmanZivari: Yes! All things work Correctly.

SalmanZivari: It's a need for a better user experience. Some users - or even many - like to have such frequently list. (I saw this request before)

SalmanZivari:
You don't want to merge it?!! it's not a good feature?!!

kjk:
1. We can't just leave all thumbnails to accumulate indefinitely. One option is to bump the number of thumbnails to some reasonably large number (128?)

2. I have to test how it works in practice before merging

SalmanZivari:
I change the FILE_HISTORY_MAX_FREQUENT to 60!! it's not any effect on number of frequently list!!!! but it can used in Clean Up Method by multiplication of 2! so no maxumum number of thumnail saved in disk is 120

SalmanZivari:
I close this conversation, but code changes remains if you want use.

Issue #855, opened 2017-08-01
Is it possible to include SublimeText in "editor_rules"? 


Romeoly:
There are so many editors included "editor_rules[]" in "AppTools.cpp": that is, WinEdt, notepad++, TeXnicCenter, TEXCNTR, WinShell, gvim, texmaker, TeXworks. 

I wonder if it is possible to include  SublimeText [https://www.sublimetext.com/](SublimeTextl) in the list? It's gathering its popularity. 

Thanks.

PR #866, opened 2017-08-17
Add Sublime Text to editor_rules

GreenLightning:
As requested by #855.

Tested with Sublime Text 2 and Sublime Text 3. If both are installed, Sublime Text 3 is used.


kjk:
Thanks!

Issue #545, opened 2016-06-02
Bookmarks: navigating like in a real book


bensoibj:
First of all Sumatra is a really nice and handy small reader. Thanks! Nevertheless, I often miss one feature which would make it behave like reading a real book.

Reading a real book, one has the ability to put a finger between the current pages whenever you want to quickly jump to another or between two or more positions in the book. This is handy when reading for instance technical documents where you often want to switch between the text part and the reference list at the end of a chapter or compare formulas on different pages, etc. Though one could "abuse" Sumatra's Favorites feature, this is not very comfortable, since too many keystrokes are necessary and no shortcut exist for jumping to a saved position.

What would be nice is if there were some, lets say 10, bookmark places, which are directly accessible through shortcuts (e.g. 1, 2, ..., 0). This would allow to easily jump across different position in the document. Whether a bookmark place is free or "engaged" could be indicated in the status or menu bar. What would be even nicer is if one of these bookmarks automatically tracked the current reading position. So if one forgets to put a finger between the current pages before jumping to another position, one could still jump back to the current reading position thanks to the automatic finger (=bookmark).

Related issues:
- #57 I cant create bookmark expediently
- #265 Add shortcut for favorites
- #119 Add bookmarks icon to toolbar.
- #470 Feature request: Button or menu choice to show the Frequently read page

sbridger:
Nice idea.
I don't know if table of contents is identifyable in a pdf, but that would be a great predefined one...

calvin-young:
This is an excellent idea and I would personally use it daily.  

Example case: I am reading a textbook and want to make a quick glance at an appendix in the back of the book.  Currently I must take note of which page I am on, and then manually punch that page number back in once I am done referencing the appendix.  It would be lovely to be able to simply flip back and forth.

nod5:
Years ago I made the helper tool [sumatra_earmarks](http://nod5.dcmembers.com/sumatra_earmarks.html) that does exactly what OP requests.

Another tool [sumatra_highlight_helper](http://nod5.dcmembers.com/sumatra_highlight_helper.html) similarly allows for quick jumps between pages with highlighted text on them.

Kevin-Mok:
@nod5 How hard would that be to implement directly into the code?

nod5:
@Kevin-Mok I don't know C++ or the SumatraPDF code so can't say how hard.

[...]

PR #869, opened 2017-08-25
Add shortcut to manually add the current page as a navigation point

ahrm:
Sometimes while reading a document I need to scroll a few pages forward or backward (without clicking on any links or bookmarks) and quickly return to where I was. Now we can do this by pressing the shortcut ('y' as of this commit) before scrolling and then return by hitting backspace.

### DIFF
--- a/src/DisplayModel.h
+++ b/src/DisplayModel.h
@@ -186,6 +186,7 @@ public:
     bool            dontRenderFlag;
 
     bool            GetPresentationMode() const { return presentationMode; }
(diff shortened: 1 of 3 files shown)


kjk:
Thanks.

While I understand this might be useful it doesn't look like it would be used by many people. For that reason I won't merge this change.

bensoibj:
I would have liked such a feature a lot. Some other users certainly too, see also #545.

Issue #619, opened 2016-09-20
PDF bookmark links not working


Sasstraliss:
I have lots of PDFs, and most of the bookmark links work fine.

Some of my PDFs have bookmarks, but the links (to the particular pages) don't work at all when clicked. They work in Adobe Reader DC though, going to the appropriate section, so it must just be Sumatra.

Here's one of the culprit PDFs (yes I own it in hardcopy too, calm down :P)... https://1drv.ms/b/s!AmLJxqfRJ4Hemkk9QlNIzplvHjs8

jiulun:
Indeed, this is a non-conforming PDF file.
If you open this file in a text editor, and search "Auto/Page". You will see find page object has /Auto/Page and pages object has /Auto/Pages. The requirement is a page or pages object shall have Type as Page or Pages respectively.
Solution 1: Search and replace "Auto/Page" as "Type/Page". This replace "Auto/Page" to "Type/Page" and "Auto/Pages" to "Type/Pages". 

Solution 2: In source code function pdf_lookup_page_number() in pdf-page.c. Relax the page type check. 
E.g.
Replace
`    if (strcmp(pdf_to_name(pdf_dict_gets(node, "Type")), "Page") != 0)
        fz_throw(ctx, FZ_ERROR_GENERIC, "invalid page object");
`
with a warning
`   if (strcmp(pdf_to_name(pdf_dict_gets(node, "Type")), "Page") != 0)
    {
        if (pdf_dict_gets(node, "Kids") != NULL)
            fz_throw(ctx, FZ_ERROR_GENERIC, "invalid page object");
        else
            fz_warn(ctx, "invalid page object");
    }
`
An addtional check is not pointing to a pages object.

Both solutios work well.

bigeast:
I also have this problem, the bookmarks work fine in Adobe Reader DC but disappear in sumatraPDF.

There is no "Auto/Page" when edit in text editor.

PR #923, opened 2017-11-05
PDF: Relax page type check (Fix issue #619)

libdor:


kjk:
Thanks for the fix!
Even better solution would be to update to latest mupdf, but that's a lot of work due to api changes.

libdor:
Do you want me to update to the latest mupdf?

kjk:
That would be great. 

Beware that this might get complicated. Not only mupdf API have changed but mupdf also uses newer dependencies (like freetype, jbig2dec etc.) so those might require updating as well. I think they also added harfbuzz as dependency, so it would have to be added as well.

We also had some custom patches against those librarries (see ext/_patches directory). Some of those might not be needed anymore but some might require re-adding).

In case you're unfamiliar with build system, see https://www.sumatrapdfreader.org/docs/Our-build-system-3114db56b1a54b9bb026da6b92d36a80.html (in short, when adding/removing files or changing compilation flags, we modify premake5.lua and premake5.files.lua and re-run premake5 to generate visual studio projects in vs2015 and vs2017 directories).

(aside: I probably will require vs 2017 soon and vs 2015 will no longer be supported).

libdor:
Finally,  can sit down to get some progress.

I have successfully compiled the mupdf 1.14.0 with the latest third party libraries, jbig2dec,  freetype2 and harfbuzz.

So,  can link the mupdf library and get the mutool. Now,  the big challenge is to fix all the issues in PdfCreator.cpp and PdfEngine.cpp, because the mupdf interface has changed so much.

Have you begun such work?  Any suggestions?

Jun

kjk:
No, I haven't done any work on that so at this point, you know more than me. I would think that changes would be mostly mechanical but as I said, I'm guessing at this point.

If you share what you have so far (e.g. by putting it on a branch in your fork) I might find time to look at it and help.

libdor:
I have pushed the changes to https://github.com/libdor/sumatrapdf/tree/mupdf-1.14.0

You may take a look and see what's the best way to integrate the changes.

________________________________

PR #950, opened 2017-12-29
Vim Commands

gustaphe:
Simply added 'h' for "left", 'l' for "right" and 'n' for "find next match".
Substituted 'a' for "turn into annotation" and (somewhat awkwardly) 'o' for next page.


kjk:
Adding 'h' and 'l' is fine. Changing 'h' to 'a' is also ok because 'h' wasn't officially shipped.

However: we can't change 'n' to 'o'. Too many people know the current shortcut, we can't change that without a really good reason/.

gustaphe:
Of course. I wasn't sure how known that command is, but I figured there was a risk.

kjk:
Thanks!

Issue #946, opened 2017-12-17
Cppcheck results


Friedrich2:
Dear developers,

I've checked the src folder with Cppcheck (version 1.61). The results are attached. Expecially the two error candidates found by the program could be of interest to you. Since Github does not support XML files, the file ending must be changed to .xml to read the file using Cppcheck.

Greetings

Friedrich2
[Cppcheck_results_SumatraPDF_xml.txt](https://github.com/sumatrapdfreader/sumatrapdf/files/1565663/Cppcheck_results_SumatraPDF_xml.txt)

Friedrich2:
One of the errors (null pointer dereference) was made deliberately to assert some condition only in debug mode. The other concerned missing square brackets behind delete <pointer to array>. The warnings are still unresolved.

GitHubRulesOK:
@kjk @Friedrich2 
Is this now closed ?

kjk:
@GitHubRulesOK No, need to fix some more issues flagged by cppcheck

kjk:
All cppcheck are now fixed.

PR #952, opened 2018-01-05
expected is pointer to array. C. f. line 79.

Friedrich2:
The function BuildTextSelList() is returning a const pointer pointing to an array. It is something to overlook easily and was found through using [Cppcheck](http://cppcheck.sourceforge.net/). It fixes the only non-deliberate error in #946. I am leaving the issue open to inspire future comitters through the - not resolved - warnings likewise mentioned in the results.xml file. Please edit my comment with respect to grammatical errors too. Many thanks in advance.


kjk:
Thanks!

PR #955, opened 2018-01-07
Perfomance improved inspired by #952

Friedrich2:
NOTE: Due to many changed function parameters (values to references!), this commit should be carefully tested.

Signed-off-by: Friedrich Spee von Langenfeld <stehlampen@arcor.de>

### DIFF
--- a/src/ChmModel.cpp
+++ b/src/ChmModel.cpp
@@ -516,7 +516,7 @@ class ChmThumbnailTask : public HtmlWindowCallback {
     CRITICAL_SECTION docAccess;
 
   public:
-    ChmThumbnailTask(ChmDoc* doc, HWND hwnd, SizeI size, const onBitmapRenderedCb saveThumbnail)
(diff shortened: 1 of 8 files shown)


kjk:
There are 3 reasons I can't merge this change.

1\. It changes too many unrelated things at once. If something breaks due to those changes, it'll be hard to figure out which change was responsible.

2\. Passing by value vs. by reference changes semantics in a subtle way. That means that each change needs to be tested at least by stepping through the code in the debugger to make sure that changed code path have been exercised.

And that needs to be tested by the person making the change because that's the bulk of the work.

3\. Most of those codepaths are not in performance sensitive code. 

cppcheck is a tool to point out areas of focus but you have to apply judgement on top of that.

Looking at the changes, those that deal with std::function arguments are no-brainer. 

Others, like Doc don't need changing. I even made explicit note in https://github.com/sumatrapdfreader/sumatrapdf/blob/master/src/Doc.h#L8 that it's ok to pass them by value.

PR #1046, opened 2018-07-22
A6 and A2 support in commandline print

baloo887:
Hi sumatrapdf :) , 
The purpose of this commit is enabing A6 and A2 paper formats, in command line print. 

I've tested the A6 with three different printers and It works fine. 

About the A2, I don't have access to a printer supporting such format, but It should work since is the exact same logic of the A6. Anyway if you think is not enough to trust it I could remove It.


kjk:
Thanks!

PR #1057, opened 2018-08-16
fix CJK line broken problem

ultranity:
use iswalnum( ) to identify CJK characters and prevent wrong new-line typesetting(actually, Chinese words break does not affect reading experience )(may have better solution)

### DIFF
--- a/src/HtmlFormatter.cpp
+++ b/src/HtmlFormatter.cpp
@@ -708,10 +708,10 @@ void HtmlFormatter::EmitTextRun(const char* s, const char* end) {
         }
 
         size_t lenThatFits = StringLenForWidth(textMeasure, buf, strLen, pageDx - NewLineX());
-        // try to prevent a break in the middle of a word
+        // try to prevent a break in the middle of a word, CJK support added. 
         if (iswalnum(buf[lenThatFits])) {
             for (size_t len = lenThatFits; len > 0; len--) {
-                if (!iswalnum(buf[len - 1])) {
+                if ((!iswalnum(buf[len - 1])) || (iswalnum(buf[len - 1]) == 256)) {
                     lenThatFits = len;
                     break;
                 }


ultranity:
fix problem described in #250

kjk:
I don't understand what iswalnum(x) == 256 means. Per documentation (https://en.cppreference.com/w/cpp/string/wide/iswalnum) iswalnum() returns either 0 or non-0.
Currently we check for 0.

ultranity:
[MSDN documentation ](https://msdn.microsoft.com/en-us/library/windows/desktop/d30dx8kc(v=vs.110).aspx/html)
 As I tested, the return value is defined in wctye.h, and it is  specified. for example, CJK character return 256, English letters return 258, digits return 4. 
this feature should be universal though I just tested it under en_US.utf8 and zh_CN.UTF-8 locale.  

anyway, this change won't affect non-CJK users and will improve the reading experience in Chinese and Japanese.
there is an example(epub). the left part is the original release version which make wrong newline after punctuations, the right part is the proper format made by my modified version, you can see the difference clearly. 
image

 if there is any other efficient way to detect CJK characters, that will certainly be a better solution. If you are not satisfied with this just-working solution, we can discuss it further.

kjk:
This is a subtle behavior that affects every file. Here's how a robust fix for this would look like:

* extract this loop into a stand-alone function, put it in StrUtil.cpp.
* add a few test cases for it in StrUtil_ut.cpp, for both cjk and non-cjk use cases, including a test that would fail with current code.
* add a fix

It seems the issue with this code is that `iswalnum` relies on current locale, which is wrong.

If there's no built-in function that returns the right response, we should write one. It's just checking number ranges. So the current equivalent of `iswalnum` is roughly:
```
bool IsCharAlphaNum(c WCHAR) {
  if (c >= '0' && c <= '9') { return true;}
  if (c >= 'a' && c <= 'A') { return true;}
  if (c >= 'z' && c <= 'Z') { return true;}
  return false;
}
```

To add CJK support, we would add the case for CJK ranges.

I don't want to use windows-specific, locale-dependent implementation that I have no visibility into.

ultranity:
Let me describe the bug in CJK text again: (1) punctuation and (2) Rich text style tag（such as \<b\>, \<span\>) will cause a wrong new line typesetting in CJK , while every Chinese character is an single word, and they are not separated by space.( A huge difference between English and Chinese( or Japanese) ). And there is no word break in CJK text reading, while sumatra have no optimization for that.

  though the reading experience in plain text format is perfect after this modification, Rich text markup can cause similar problems with incorrect newline typesetting.
another example with screenshot and I highighted where wrong newline format exists.
[test - test.zip](https://github.com/sumatrapdfreader/sumatrapdf/files/2298266/test.-.test.zip)


image

  Actually, I have not fully understand how sumatrapdf handles those complicated texts, so now I don‘t know how to fix it, need your help @kjk

kjk:
Thanks for detailed information and test document.
I hopefully made a fix in https://github.com/sumatrapdfreader/sumatrapdf/commit/33fcbc31d7ccbedd66c049d1117049b29306876e
Please see if it fixes it for you either by compiling latest code or using latest pre-release https://www.sumatrapdfreader.org/prerelease.html

ultranity:
This does solve the previous problem. 
But If possible, I hope the second problem can be solved too. However, I don't understand why those Styled text will cause similar problem  (shown in the left part of the second screenshot), do you have any idea on it?

ultranity:
sorry for my carelessness. I use the same test epub file as before, Line 1, 3, 9, 11 break in Sumatra pdf.
the html code:
image
 the correct style rendered with sigil compared with sumatra pdf:
image

GitHubRulesOK:
Oddly enough other github projects are trying to address similar issues OCR gImageReader has a commit for their CJK space issues at https://github.com/napasa/gImageReader/commit/30f24db91bb15440c2b41f24aa6ee3b8098a99da

kjk:
Good catch. It's also the same solution.

GitHubRulesOK:
Must be Karma since I was writing up my notes on OCR at the time https://github.com/GitHubRulesOK/MyNotes/raw/master/AppNotes/OCR%26R2V-InProgress.pdf

PR #1060, opened 2018-08-20
Don't sort again with inferior sort.

glachancecmaisonneuve:
Files were already sorted in ImageDirEngineImpl::LoadImageDir.  Sorting again with strcmp will undo the natural sorting.


kjk:
Thanks!

PR #1063, opened 2018-08-24
Fix compilation when using the Windows 10.0.17134.0 (RS4) SDK

Mattiwatti:
Hi,

Thanks very much for your work on SumatraPDF!

PR/commit is as per title and should be straightforward. This struct and the value of `ProcessImageLoadPolicy` were moved fom the MSDN documentation to `winnt.h` in the Windows 10 RS4 SDK, which causes a compile time error because these are now redefinitions of existing types.

No functional change intended.

### DIFF
--- a/src/utils/WinDynCalls.cpp
+++ b/src/utils/WinDynCalls.cpp
@@ -253,6 +253,8 @@ void NoDllHijacking() {
     }
 }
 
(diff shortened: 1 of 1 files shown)


kjk:
Thanks!

PR #1064, opened 2018-08-24
Do not load more than one version of comctl32.dll

Mattiwatti:
The current version of SumatraPDF (git, not 3.1.2) has a bug that causes `comctl32.dll` to be loaded twice with different versions:
common_indeed
This happens on all supported versions of Windows, from XP through 10.

The cause of this is the well-intentioned but somewhat misguided `NoDllHijacking()`, which attempts to load `comctl32.dll` from the `system32` directory using `SafeLoadLibrary`. However, at this point (and in fact, before `WinMain` is ever reached) `comctl32.dll` v6.x has long since been loaded. This is because of the explicit DLL dependency declared in the application manifest. You can verify this using a debugger such as WinDbg or x64dbg that stops at the system breakpoint.

Complicating the matter is the fact that `comctl32.dll` has been declared as a delay-load dependency. This means that the ntdll loader cannot be sure which, if any, additional version(s) of `comctl32.dll` may be requested at runtime, because it is possible to dynamically activate and deactivate Fusion (SxS) contexts.

So this commit consists of two parts:
1. Change `comctl32.dll` from a delay-load dependency to a static import.
2. Remove `comctl32.dll` from the list of DLLs to 'preload' (although 'post-load' would actually be a better description in this case :wink:)

You may wonder whether this does not open up the possibility of a DLL preloading attack. The answer to this is no, because the side-by-side resolver will reject DLLs that (a) do not match the manifest dependency declaration, or (b) are not located in either `system32` or the `WinSxs` folder (which has the same security level as `system32`).

### DIFF
--- a/src/utils/WinDynCalls.cpp
+++ b/src/utils/WinDynCalls.cpp
@@ -241,7 +241,7 @@ HRESULT GetReservedNotSupportedValue(IUnknown** punkNotSupportedValue) {
 }; // namespace uia
 
 static const WCHAR* dllsToPreload =
-    L"comctl32.dll\0gdiplus.dll\0msimg32.dll\0shlwapi.dll\0urlmon.dll\0version.dll\0windowscodecs.dll\0wininet.dll\0\0";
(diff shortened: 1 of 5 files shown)


kjk:
Thanks!

PR #1071, opened 2018-08-31
Remove redundant call to CloseFigure. Keep it consistant with define …

zhaohui0923:
According to the MSDN, there is no need to call CloseFigure() after the AddRectangle() or AddEllipse() call;
and in the same function, 10 lines below, when the original code use "AddEllipse" to draw the 'x', there is no call to the CloseFigure(). So, for efficiency and consistancy purpose, I removed the redundant CloseFigure() call.


kjk:
Thanks!

PR #1073, opened 2018-09-04
Add handle validation check before close

zhaohui0923:
According to the MSDN, CloseHandle() "Closes an **open object** handle", and in the remark section it says "If the application is running under a debugger, the function will throw an exception if it receives either a handle value that is not valid or a pseudo-handle value. "，So I add a validation check before close it to make the code more stronger.

Also, when implement the RAII, maybe it's better to implement the "NonCopyable" paradigm for some wrapper, It's zero cost and can make the RAII wrapper much more harder to be misused, It's just a suggestion :).

### DIFF
--- a/src/Tabs.cpp
+++ b/src/Tabs.cpp
@@ -90,7 +90,6 @@ class TabPainter {
         GraphicsPath shape;
         // define tab's body
         shape.AddRectangle(Rect(0, 0, width, height));
-        shape.CloseFigure();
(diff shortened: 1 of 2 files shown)


kjk:
Thanks.

Re: non-copyable. Sure. Most of this code was written before non-copyable was a well known pattern and before C++-11 was a thing. There's plenty things that could be improved.

Issue #1059, opened 2018-08-19
Chinese text shouldn't break on <b> etc.


kjk:
`foo<b>bar</b>foo`, if it were cjk text, shouldn't break on `<b>`.
See https://github.com/sumatrapdfreader/sumatrapdf/pull/1057#issuecomment-414105612 for more details.

GitHubRulesOK:
Sorry crossed edit I think this is not a separate issue unless a problem in non CJK

ultranity:
> I think it might be rather subtle/involved and related to how html is parsed and laid out.

there is no space between CJK  words or sentence unlike English, this difference caused the problem when htmlparser try to handle CJK sentences as words since its usually too long to put all of them in one line.

kjk:
should be fixed by https://github.com/sumatrapdfreader/sumatrapdf/commit/158b13ce7bdb4e80cfdc0822336134042fbd26f7

PR #1107, opened 2018-12-23
tweak HtmlFormatter::EmitTextRun to get better layout for CJK 

ultranity:
fix #1057 and  #1059 to get better layout for CJK


kjk:
thanks!

PR #1206, opened 2019-09-16
update translations

sebras:
I happened to look at sumatrapdf source today and I saw that the latest commit was "update translations". This prompted me to take a quick look at the Swedish translations, and so here are my suggested changes. :)


kjk:
Thanks but the way Sumatra's translations work is that the source of truth in is in http://www.apptranslator.org/app/SumatraPDF

The file you modified are over-written is auto-generated from data downloaded from apptranslator.org

Could you make this changes in http://www.apptranslator.org/app/SumatraPDF/sv ?

sebras:
> Could you make this changes in http://www.apptranslator.org/app/SumatraPDF/sv ?

Sure, done! :)

PR #1219, opened 2019-10-22
Fix bugs and rename func

ShadowZzj:
The old DisableDataExecution() intends to "EnableDataExecution). It desires a change.

(Print.cpp)I have tested the code near devMode. I find that there is an exception in the program when executing the second DocumentProperties(). As I observe, the reason is the buffer is not enough.  I find that the structSize returned by DocumentProperties() is above 4000(driver-specific). However, sizeof(DEVMODE) is below 100. This leads to an overflow. 

(Vec.h) StealData() will set the size to zero, which leads to that this->size() always return 0. That is excatly what we want.

### DIFF
--- a/src/EngineDump.cpp
+++ b/src/EngineDump.cpp
@@ -425,7 +425,7 @@ int main(int argc, char** argv) {
     UNUSED(argc);
     UNUSED(argv);
     setlocale(LC_ALL, "C");
-    DisableDataExecution();
+    EnableDataExecution();
 
     WStrVec argList;
     ParseCmdLine(GetCommandLine(), argList);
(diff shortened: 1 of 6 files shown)


kjk:
FYI: I fixed the sizeof(DEVMODE) in https://github.com/sumatrapdfreader/sumatrapdf/commit/0128ff1339a0968002bc60d4cc9769d91fbe6f71
Thanks for pointing this out.

StealAsOwnedData() was correct because StealData() would call strlen but your version is better so I've made a similar change in https://github.com/sumatrapdfreader/sumatrapdf/commit/228b7f0d7f1cd92e6d9a892ab921a044f5630086

DisableDataExecution() is the right name as it disables the execution of code in non-executable pages by enabling DEP (Data Execution Prevention). It's a bit confusing.

PR #1220, opened 2019-10-28
Fullscreen & rotation bindings from Ds886's fork

ImportTaste:
From a fork by @Ds886, seems useful to add officially. Fullscreen toggling is `f`, page rotation is `[` and `]`. Not much more to say.

### DIFF
--- a/src/SumatraPDF.cpp
+++ b/src/SumatraPDF.cpp
@@ -3476,6 +3476,18 @@ static void FrameOnChar(WindowInfo* win, WPARAM key, LPARAM info = 0) {
         case 'z':
             win->ToggleZoom();
             break;
+        case '[':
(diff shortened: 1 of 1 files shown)


kjk:
Thanks. Not sure if I'll keep 'f' long term (a bit easy to hit for a significantly large change in the UI). Let's see how it works in practice.

Ds886:
Sorry if it is late, I don't use git as much and just now found out how to interact with your merge.

I have a somewhat vim binding habits so it made sense because of that. for more standard usage I would add ctrl or alt so it wouldn't be accidentally pressed(In mupdf there is similar issue with the presentation mode that really annoys me).
 alternatively it could be  made user configurable although from looking at the code it might require more elaborate implementation

PR #1223, opened 2019-11-02
Added search functionality from selection

8bitMafia:
Simplistic ease of access addition. I find it annoying when programs don't include a "search with google" functionality because it is so much faster. Included is a hotkey and an option to change the search engine based on URL.

### DIFF
--- a/src/Menu.cpp
+++ b/src/Menu.cpp
@@ -121,6 +121,7 @@ static MenuDef menuDefView[] = {
     { SEP_ITEM,                             0,                          MF_REQ_ALLOW_COPY | MF_NOT_FOR_EBOOK_UI },
     { _TRN("Select &All\tCtrl+A"),          IDM_SELECT_ALL,             MF_REQ_ALLOW_COPY | MF_NOT_FOR_EBOOK_UI },
     { _TRN("&Copy Selection\tCtrl+C"),      IDM_COPY_SELECTION,         MF_REQ_ALLOW_COPY | MF_NOT_FOR_EBOOK_UI },
+    { _TRN("Search Selection\tCtrl+Z"),     IDM_SEARCH_SELECTION,       MF_REQ_ALLOW_COPY | MF_NOT_FOR_EBOOK_UI | MF_REQ_INET_ACCESS },
 };
 //] ACCESSKEY_GROUP View Menu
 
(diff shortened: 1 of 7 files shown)


kjk:
Thanks.

I need to think about it. 

Part of the appeal of Sumatra is simplicity, which includes trying to be minimal about number of menu items etc.

Given that an alternative is Ctrl-C, Alt-Tab to the browser, Ctrl-V and Enter, I'm not sure if this functionality out-weights the cost of adding additional menu item.

I would be more inclined to add more functionality if Sumatra had a concept of comand palette (like e.g. Ctrl-Shift-P in Visual Studio Code), which allows to add large number of additional comments without cluttering the primary UI.

8bitMafia:
Sure, I was going to use it for my own purposes, but you could add that command palette. Or, you could make some sort of loader for python scripts that would allow for modification much like other programs have plugins, but I'm not sure how much people want to change up their experience lol.

kjk:
Re: Python. I was actually thinking JavaScript, using either mujs or quickjs. Those are light-weight interpreters that can be shipped with Sumatra (Python is rather large).

I think it would be possible to (re)write a large portion of the app in JavaScript (by adding bindings to win32 apis) and use it as a way for easy customization, like you've said, by loading user scripts.

Shakil-Shahadat:
I would like to have the search functionality on the right-click menu. Being able to add multiple search engine will be even better. Like, one for Google, one for oxford dictionary.

GitHubRulesOK:
I second the idea of calling a command with current selected textual string
At present this could be simply the current selection then later this could possibly be extended to different calls per hotkey combo

Can I suggest it be left to the user to decide the format of the call
This is because I uses multiple browsers for multiple functions (none are the default http browser)
so at its simplest there is ability to add a "prefix" and "appendage" perhaps read from the settings.txt file ?

This could in effect be similar to the current call editor with "filename" / "page number" just adding  "content string" note C was reserved for column but never functional so best to use S or $ for string

Thus one user can have a google search whilst another can use grammerly look-up and another can call Firefox portable query

kjk:
I don't want to keep a PR open as I don't plan on merging this soon.

I've opened an issue https://github.com/sumatrapdfreader/sumatrapdf/issues/1280 that references this PR, so it doesn't get forgotten.

kjk:
I've added such functionality in https://github.com/sumatrapdfreader/sumatrapdf/commit/d710475e28acc407cb7fa6fc08e00d77ef8753a9

PR #1266, opened 2019-11-26
Change default color value of MainWindowBackground

Bambooin:
If you open portable sumatrapdf in a new folder, the default value in MainWindowBackground is:
MainWindowBackground = #80fff200
And the the color will be setted with last six char fff200 and ignore 80.

Anyway #80fff200 is not match in settings3.2.html

Last line change is due to file not end with /n


kjk:
As far as I can tell it's parsed correctly (see SettingsUtil.cpp). It might not take the desired effect, but it's a different issue.

Issue #162, opened 2015-04-08
Closing window in tabbed mode by double-clicking menu button


DuMOHsmol:
Hello, everyone!

I've got a small feature request about adding an option to close main window with double click on menu button. Most programs have an icon in top left corner and they can be closed by double clicking that icon. It's a little-known, but very convenient feature -- you can close windows using both sides of their titles. But SumatraPDF has a menu button there and when I try to close its window from left side of the title, I just open and close menu. That's not a big issue, but just is a little confusing.

Thank you for your attention :)

DerpMcDerp:
Right clicking on that button should display a "Restore | Move | Size | Minimize | Maximize | Close" popup.

rottweiller76:
hey, this is a very nice thing that is missing :( i always close the programs with double clock top left 

as you can see in the picture the option to close is very very hidden when compared to almost all other programs which react at double click with - instant close / alt F4 
or with single they open the close option directly not hidden in other menu

![double click top left](https://cloud.githubusercontent.com/assets/12532841/7731895/6e1c9d66-ff2c-11e4-8e98-506223a98157.jpg) 

please make this happen, i love the program and this is the only issue i have with it :) 

ps. i made this account just to add this comment

Chungalin:
I was coming to fill in a bug, but since this is the same I was about to report, I just second this proposal.

This is a general usability issue in Windows. Any program that uses standard windows API allows to be closed with a double click on window icon. I find very annoying when a program customizes title bar in a cool way but forgets to support all standard functionality.

Also, it’s about two lines of code!

Please consider it.

petrasteri:
In my opinion, it might not be a good idea to make "double-click on menu button = close window", or the interval to accept this behaviour has to be very short.

People would hate it because simple mistake clicking on menu button and close it = closing window!

On the other hand, putting menu button on the left is rare in desktop app.  Other modern software without menu bar like Firefox, Google Chrome, have the menu button on the right.  Examples of menu button on the left I can find are Calculator (provided in Win10), JavaScript Studio (downloadable from Store) and some other "tiled" apps provided in Win10.  In none of them, double-click on menu button would close the window.

rottweiller76:
97% of windows apps have that double click close there:) 

    On Tuesday, August 22, 2017 12:26 PM, 石庭豐 <notifications@github.com> wrote:
 

[...]

Issue #1053, opened 2018-08-03
titlebar hard to grab when many tabs open


e4711s:
In tab mode the tabs reach into the titlebar, leaving only a couple of free pixels above the tabs to grab the window.

GitHubRulesOK:
That is unusual 
you possibly know thay you can from the pull down menu use Window > Move then any arrow key will allow you to move the window.
As to cause, it may be related to screen dpi can you confirm resolution of screen AND what percent your working to, usually found by right click on desktop then select display settings ?

SumatraPeter:
> That is unusual

@GitHubRulesOK: Here's the 'grabbable' area of Sumatra's titlebar, which admittedly is a bit narrow:

![image](https://user-images.githubusercontent.com/15380332/43733928-d4d992ee-99d3-11e8-997f-35adef46a414.png)

Perhaps the tabs could be redesigned to take up the _entire_ height of the titlebar as in the case of Firefox, but leave clear gaps on both sides to aid in grabbing and moving the window with a mouse.

![image](https://user-images.githubusercontent.com/15380332/43734051-2a3f9ea4-99d4-11e8-9bfa-13f0d7998c58.png)

Note a clever aspect of Firefox's design, in that the empty area to the left of the tabs disappears when the window is maximized (the first tab starts right at the top left of the window in that case).

Various tab-related improvements have been on my wish list ever since Sumatra started supporting tabs (see [here](http://forums.fofou.org/sumatrapdf/topic?id=3184717) and linked GitHub issues), but unfortunately there's no-one around to implement any of them. 😞

GitHubRulesOK:
Peter whilst I agree there are many requests/improvements needed for tabs
I see the area above yours is about 14 pixels rather than the "couple" (two?) as mentioned by @e4711s hence the `That is unusual` 
I have on occasion mis-grabbed a tag and moved it rather than the whole window but I put that down to my haste or bad eye co-ordination.
Personally I think the current tabs look ok when maximised and (apart from color/logo) not that different to your Firefox example (which seems to have a space top left)  and I would rather find the tribar options button in the traditional top left menu position well away from close / resize etc at top right see my screen
![untitled1](https://user-images.githubusercontent.com/12754913/43741592-6f18ee10-99be-11e8-8e63-b3790d75c826.png)
[...]

Issue #423, opened 2016-01-26
Feature Request: Add one or two pixels above tabs in title bar when maximized


voidray:
In Windows 7 and above it is possible to drag a window to and from the top of the screen. When sumatra pdf is maximized and many tabs are open it is not possible to click and drag the window by the title bar. Instead the tab is clicked.
It would be great to have that one or two pixels (or whatever is appropriate) to be able to support this scenario. 

The same problem exists in firefox by the way. I use the addon "Extra Padding When Maximized" in that case.

stefanocaselli:
Please make it optional: I like to move the mouse to the top of screen and just click the tab: if you add some extra pixels at the top then I have to position the mouse pointer more accurately.
By the way: also in Firefox I don't want some extra pixels at the top of the screen :-)

voidray:
Thanks, yes an option would be perfect.

timhandschack:
+1 for this feature request
But instead of adding some extra pixels, please consider either 
following the (old) OS constraints of having a title bar with no content apart from min/max/close buttons or 
simply not filling the complete title bar with tabs (but leaving some padding towards the buttons).

FrankSFLYS:
Another option is to add more space to the right of title bar. Considering opening a lot of tabs, it is difficult to drag the window. So just do as what chrome does now, no matter how many tabs opened, there is a wide fixed area on the right so that we can drag the window without clicking on the minimize button.
![Snipaste_2019-09-14_22-28-33](https://user-images.githubusercontent.com/10843159/64909501-a6a12180-d73f-11e9-9223-87c6296853a2.jpg)

SumatraPeter:
@FrankSFLYS: That is precisely what I had suggested quite some time ago as well: https://github.com/sumatrapdfreader/sumatrapdf/issues/1053#issuecomment-410807509

GitHubRulesOK:
@kjk this issue could be consolidated with #1053

kjk:
Should be fixed thanks to @amurzeau adding app icon in the top-left menu bar that can be used to move the window and is always there.

PR #1372, opened 2020-01-03
add option to exit when double cliking on the menu button

amurzeau:
This commit add a boolean option MenuButtonDblClickExit.
When true, a double click on the menu button will cause SumatraPDF to
exit (the menu button is only shown when UseTabs is true).

This allows to close SumatraPDF like other Windows application using a
double click on left top icon of the title bar.
The double click duration is the one configured in Windows' settings.

Issue: #162

### DIFF
--- a/src/Caption.cpp
+++ b/src/Caption.cpp
@@ -16,6 +16,7 @@
 #include "Translations.h"
 #include "resource.h"
 #include "Menu.h"
+#include "GlobalPrefs.h"
 
 using namespace Gdiplus;
 
@@ -166,6 +167,12 @@ static LRESULT CALLBACK WndProcCaption(HWND hwnd, UINT message, WPARAM wParam, L
                         SetTimer(hwnd, DO_NOT_REOPEN_MENU_TIMER_ID, DO_NOT_REOPEN_MENU_DELAY_IN_MS, nullptr);
                     }
                     SetFocus(win->hwndFrame);
+                }
+            } else if (win && BN_DBLCLK == HIWORD(wParam) && gGlobalPrefs->menuButtonDblClickExit) {
+                WORD button = LOWORD(wParam) - BTN_ID_FIRST;
+                if (button == CB_MENU) {
+                    // Menu button double clicked, exit now
+                    SendMessage(win->hwndFrame, WM_CLOSE, 0, 0);
                 }
             }
             break;
--- a/src/SettingsStructs.h
+++ b/src/SettingsStructs.h
@@ -372,6 +372,9 @@ struct GlobalPrefs {
     bool showStartPage;
     // if true, documents are opened in tabs instead of new windows
     bool useTabs;
+    // if true and useTabs is enabled, double clicking on the menu button will exit
+    // (to match windows behavior when double clicking on system icon on the left of the titlebar).
+    bool menuButtonDblClickExit;
     // information about opened files (in most recently used order)
     Vec<FileState*>* fileStates;
     // state of the last session, usage depends on RestoreSession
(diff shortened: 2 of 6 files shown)


kjk:
I've been thinking about it and here my thoughts.

Sumatra's philosophy is to keep UI simple and number of settings to minimum. In this case, it either always works this way or never.

However, I don't want double-click on menu icon to trigger this. It's too easy to do accidentally.

Would you be willing to do a little bit more work and implement emulating standard windows behavior by adding sumatra icon before the menu and implement standard behavior on the icon (click to show the system menu, double-click to exit)?

That way Sumatra would behave like standard apps and while in general I want to maximize the amount of space used for content, an icon doesn't take much space.

amurzeau:
Ok, so as I understand, it would be a permanent solution (without option to enable or disable) ?

So instead of this (don't mind the yellow-ish titlebar):
image

It will be like this (made using mspaint :)):
image

In that case, I can try to do this for sure.

kjk:
Yes, exactly that.

kjk:
Implementation-wise, maybe all it's needed to implement the behavior is to respond to WM_NCHITTEST with HTSYSMENU for the area with the icon? Just thinking out loud.

GitHubRulesOK:
~interesting that apps like firefox have other options such as move in addiion to a second quick click is close by default

image

SumatraPeter:
Personally, anything that reduces space for tabs is a negative for me.

> ~interesting that apps like firefox have other options such as move in addiion to a second quick click is close by default

@GitHubRulesOK: In FF (whether maximized or not) I can't even bring up the system menu with a left click and only Alt+Space does the trick. How are you managing "a _second_ quick click [to] close by default" when even the first doesn't seem possible?

GitHubRulesOK:
@SumatraPeter My firefox is  v71portable and thus I often note differences to others mainstream screenshots [may also explain why attempts at using spellcheck extensions usually fail] but If I left click slowly I see the context menu (same as right click) and if i click quickly enough it is accepted as a double left click

GitHubRulesOK:
@amurzeau & @kjk 
If draging the icon can move the whole window then it would also address https://github.com/sumatrapdfreader/sumatrapdf/issues/1053#

SumatraPeter:
@GitHubRulesOK: First time I'm ever hearing of a slow left click behaving any differently from a normal speed one! 😆 Doesn't seem to work on FF71.0 64-bit installed version here. I'm assuming the window's not maximized and you're left-clicking in the blank area to the left of the first tab?

GitHubRulesOK:
@SumatraPeter there is NO space to the left of my firefox v71x64 tabs (nor draging the space to right has any effect) possibly due to a customise options. Tabs are the second row down and it is the top title bar that has the program icon (firefox logo) I can move the window by dragging that title bar but NOT drag that far left icon to move the window it stays fixed.
However on first left press of mouse on the icon I am offered the context menu shown and if i click a second time within the doubleclick time limit THEN it actions the closure of firefox

SumatraPeter:
> there is NO space to the left of my firefox v71x64 tabs (nor draging the space to right has any effect) possibly due to a customise options. Tabs are the second row down and it is the top title bar that has the program icon (firefox logo)

Ok, so it does sound like you have some sort of customized setup because I haven't seen the traditional titlebar with app icon to the left in FF for years now. Perchance have you set `browser.tabs.drawInTitlebar` to **true**? For me personally that's just a waste of vertical screen space. Might as well use the titlebar for tabs as well. Anyway, if you have the FF icon showing up then of course it'll work the traditional way to bring up the system menu, and to close the app on double-clicking.

If this is replicated in Sumatra then I hope there'll be a way to toggle it in Advanced Settings also for those of us who don't like unnecessary space wastage. In any case I tend to run Sumatra maximized, so for me it's far simpler to just blindly toss the cursor to the top right corner of the screen and click once to close the app. That or good ol' Alt+F4 is _far_ faster and easier any day IMO than seeking out the system menu icon to double-click in order to close the app.

kjk:
Closing this pull request. Hopefully there'll be a tweaked version as discussed above.

amurzeau:
I've managed to have something working, with both double clic to close and drag to move behavior.
I did not use NCHITTEST because the window does not receive it as there is a button (which take mouse event instead).
But that would not allow to implement "drag to move the window" anyway.

The solution I've come up does this:
- Add a button before the menu button in the caption
- React to WM_MOUSEMOVE when left clic is pressed to trigger the move window behavior (#1053 and #423) (Using `SendMessage(win->hwndFrame, WM_SYSCOMMAND, SC_MOVE | HTCAPTION, 0);`, I trigger the exact same behavior as when the title bar is dragged (including move to screen top to maximize and so on)
- React to WM_LBUTTONUP (which is send to the button only when not having dragged the window) to display the GetUpdatedSystemMenu
- React to WM_LBUTTONDBLCLK to close the main window
- Draw the main window icon in the new button as would do Windows (I've tested high DPI settings and it works fine)

Need some more testing tomorrow and I will be able to push that code soon.

kjk:
Sounds good!

PR #1410, opened 2020-01-15
caption: add system icon to exit when double cliking and move window on drag

amurzeau:
This commit add a new button in the caption bar as the system icon.
This allow to close the application on double click and to move the window
when dragging the button (useful when the titlebar is otherwise full of
tabs).

That new icon use the main window icon as would do Windows.

Issues: #162, #1053, #423
Related: #1372


amurzeau:
Here are some images of the result (the icon is the first one in the title bar on the left):
image

GitHubRulesOK:
Looks like it will help tablet users get a good strong point to focus for moving without risk of hitting minimize

kjk:
Thanks, looks good!

PR #1513, opened 2020-03-21
fix(doc): fix profile path close #1512

### LINKED ISSUE #1512: Documentation bug due to move of Settings file
As we've discovered, in v3.2 settings file is no more in %APPDATA% but rather in %LOCALAPPDATA%

But documentation (in this page : https://www.sumatrapdfreader.org/settings/settings.html ) has not been updated, and still refers to %APPDATA%. This is confusing.

BTW, it would have been great to automatically move existing settings file at 3.2 installation over an existing install, or at least to inform users of the move and ask hem to do the move manually...
While this move is not even mentioned in the version history....

Bambooin:


kjk:
Thanks!

PR #1592, opened 2020-05-26
Updated icons to a modern design. 

cndjonno:
I created new icons to match the trend of flat modern design.  and improve consistency across sumatras' icons. I hope you enjoy them.

Previous icons moved into a folder in gfx for reference.

Old vs New


GitHubRulesOK:
Just a coment about the yellow
WatchmenPDF


and ensuring 16 pels is workable at 400%
image

cndjonno:
I updated the yellow color as per the reference picture and also updated 16x, 32x, and 48x for for better readability by maximizing the font.

kjk:
Thanks. Those are good looking icons.

However:

I really need the source material in Figma so that I can create derivatives in the future. I can handle creating .ico files.

I've imported the .xd into https://www.figma.com/file/2nB0lLoB4V0qzik4QAWROQ/Sumatra-icons-imported-as-svg?node-id=0%3A1 but it's not great. I had to export as .svg because figma doesn't support importing PDF or .xd.

For reference, current icons are in https://www.figma.com/file/2JsWwIZWlWEPJ1YwwLkusj/Sumatra-icon-(Copy)?node-id=0%3A1 and https://www.figma.com/file/P5DWXO4ZG73B9ATNjYPhvGgD/Sumatra-icon

Also, I'm not sure if I want to update icons. Maybe just the file association icons.

cndjonno:
> I really need the source material in Figma so that I can create derivatives in the future. I can handle creating .ico files.

Good to know. 

> Also, I'm not sure if I want to update icons. Maybe just the file association icons.

I understand the concern which is why I updated the icon to be quite similar to the original including all the same text and very close in color.  I essentially moved to top left to the bottom right in order to make it appear like multiple pages behind the icon in order to give it a real material look and to give more instant recognition to people who could be new to Sumatra and what it does. However since it literally has PDF in the icon, I doubt that is a concern over all.

Either way I appreciate this open source program and wanted to contribute back.

GitHubRulesOK:
@kjk 
I think it would help any artist to have a clear scope of work
Are you accepting / rejecting the idea of curled pages in bottom right ?
Also the range of icons needed here is a hints sheet for discussion
image

cndjonno:
> I really need the source material in Figma so that I can create derivatives in the future. I can handle creating .ico files.

@kjk Out of curiosity, why do you need the source in Figma?

kjk:
Figma is free and available via the web browser so I (and anyone else) can use it easily.

cndjonno:
> Figma is free and available via the web browser so I (and anyone else) can use it easily.

Yeah makes sense.  I only choose Adobe XD over Figma back in 2018 due to Figmas' limitation on the number of projects since they both are free.  

Did you have anything specific you would like changed on the icons?  

I also updated the PDFdoc icon to be less detracting of the text on the page and added SVG folder for easier importing into Figma.

Bambooin:
Pretty looking icon for SumatraPDF, waiting for approve and test.

kjk:
Thanks, I've added the icons as new files in https://github.com/sumatrapdfreader/sumatrapdf/commit/9dd0db6d9d602698b5829185733581cc3cf39494

vlakoff:
With all due respect, I’m fine with the new icon set overall, but the second icon throws me off — it resembles a redacted Epstein file more than a regular file.

vlakoff:
I noticed in the [committed files](https://github.com/sumatrapdfreader/sumatrapdf/commit/9dd0db6d9d602698b5829185733581cc3cf39494#diff-208101738f8bc20896f2ad544f14dca23dc25b792b80570f51812fa56e250ce0) that alternative variants were provided for this particular icon, and I saw the [PdfDoc – 1.svg](https://github.com/sumatrapdfreader/sumatrapdf/blob/9dd0db6d9d602698b5829185733581cc3cf39494/gfx/gfxalt/svg/PdfDoc%20%E2%80%93%201.svg) variant, which I think goes in the right direction and looks much better.

vlakoff:
There is still something I don’t fully understand: what is the purpose of the 2nd versus the 3rd icon in these sets? Both appear to be “PDF filetype association” icons, so there seems to be redundancy.

My guess would be the following — please correct me if I’m wrong:
* In the previous icon set, there was still the original, legacy PDF icon.
* Later, a unified set of file‑association icons was created, including a new PDF icon, which resulted in having **two** different PDF‑related icons in the tree.
* Then, when the “new” icon set was created in this PR, both of those icons were simply redesigned, preserving the same duplication.

Is this the correct explanation?

vlakoff:
FWIW, I played around a bit with icon creation and ended up with these results for the file‑association icon:

Regular icon:
SumatraPDF alt icon (regular)

Simplified icon for 16×16 display:
SumatraPDF alt icon (small display size)

Not sure I would pick them over the current ones, but I figured I’d share the experiment.

PR #1622, opened 2020-06-14
Show information icon instead of exclamation warning, if no updates available

dan2468:
When checking for updates, the warning icon is shown in the message box if there are none.

A calmer information icon should be shown if the user already has the latest version.


kjk:
Thanks!

Issue #1307, opened 2019-12-08
Suggestion about the interaction of RH scrollbar


GitHubRulesOK:
A fairly large number of requests/issues arise due to different user requirements from the scrollbar.

A potentially theoretic solution is to have three states for ALL cases via two options

1) some users do not want the scrollbar showing in all modes such as continuous pages in fullscreen so as to maximise width, the feature request has generally been an icon or menu option to "Not show scrollbar" 
? that might be fairly easy as simply ON/OFF to implement (but I guess would affect many screen width code locations) ?

2a) Other users want the scrollbar to appear and show the position in current single page generally as it is now when zoomed in etc.
2b) Yet other users, even when in say "fit page to width" want the scrollbar to appear and behave as it would do for all pages in a document in "continuous page" mode (i.e. document vs page mode)
This would minimally affect current page zoomed to say fit width (as requested for mixed page size comic books) and likely require code adjustment to over-ride the page only scrollbar plus require overidding the page limit by applying an automatic "page Up/Dn " to auto switch pages during scroll or up/down arrows etc.

If these two settings were achievable it should answer most requests for fixing single page size whilst enabling/disabling a continuous scroll bar.

GitHubRulesOK:
With scrollbar setting persisted most related issues should also be addressed thus am closing this mixture of issues as primarily completed.

Issue #1211, opened 2019-09-26
[Feature request] Hide scrollbar in fullscreen mode


rotvel:
Like Adobe Reader hide the scrollbar in fullscreen mode, even if there's scrollable content. 

It's distracting with a scrollbar when immersing in a book or a comic. 

Should probably be an optional feature.

SumatraPeter:
Here's the [developer's response](https://forum.sumatrapdfreader.org/t/add-option-to-hide-scrollbar/966/3?u=sumatrapeter) on the forum to a similar request.

GitHubRulesOK:
SumatraPDF in F5 (single page) mode was introduced as comparable / similar to Adobe Reader Full Screen mode, F11 is an alternative for switching to Fullscreen If you do that in non scrolling state (i.e. single pages) then the scrollbars are "hidden", so I am closing this as it has an existing equivalent feature.
The scrollbars are only shown when required !

GitHubRulesOK:
Hiding scrollbars for non fullscreen has now been added to the current daily pre-release and should be available for most presentation modes including pictures
![image](https://user-images.githubusercontent.com/12754913/86125732-0ac98200-bad5-11ea-8512-101ec2b5f185.png)

lukewma:
Hi, sorry to touch on this again, but is there a reason why the settings option is disabled? I'm using the 3.3 preview, and while the scrollbars can be disabled on a per document basis, they come back on all documents open every time i open a new instance. I was looking through the issue, and maybe i'm misunderstanding something here, but is there a reason why the user setting commit [https://github.com/sumatrapdfreader/sumatrapdf/pull/1647/commits/b7e91e35ae1de66450955faab5091d18132ab2cf](url) isn't used?

GitHubRulesOK:
"scrollbars can be disabled on a per document basis, they come back on all documents open every time i open a new instance."

AFAIK they stay **OFF throughout the session** and also are respected opening another non tabbed instance or tabbed instance  Via CTRL N with/without shift

Only a restart or new session reverts to the initial default behaviour

PR #1647, opened 2020-06-29
Add option to hide scrollbars for PDFs

YongJieYongJie:
I added an option to hide/show scrollbars for my own use, and thought I'll share it upstream. If this feature is useful, do let me know how the code could be tidied up before merging in.

### Details:
 - Added `HideScrollbars` to the settings, under `FixedPageUI`
 - Scrollbars will be hidden / shown for PDFs based on the option
 - Works for fullscreen single page continuous mode, also tested on the modes that are accessible via shortcut keys Ctrl+1 / 2 / 3 / 5 / 6 / 7 / 8 / 0

### Demonstration:
toggling-scrollbars-visibility

and with overlap showing key-presses:
toggling-scrollbars-visibility-with-keys-overlay

### Previous discussions:
 - #1307
 - #1211 
 - https://forum.sumatrapdfreader.org/t/add-option-to-hide-scrollbar/705/5?u=sumatrapeter

### DIFF
--- a/src/Canvas.cpp
+++ b/src/Canvas.cpp
@@ -1182,6 +1182,12 @@ static LRESULT WndProcCanvasFixedPageUI(WindowInfo* win, HWND hwnd, UINT msg, WP
         case WM_GESTURE:
             return OnGesture(win, msg, wp, lp);
 
+        case WM_NCPAINT:
+            if (gGlobalPrefs->hideScrollbars) {
+                ShowScrollBar(win->hwndCanvas, SB_BOTH, FALSE);
+            }
+            return TRUE;
+
         default:
             return DefWindowProc(hwnd, msg, wp, lp);
     }
--- a/src/Canvas.cpp
+++ b/src/Canvas.cpp
@@ -1330,7 +1330,7 @@ static void OnTimer(WindowInfo* win, HWND hwnd, WPARAM timerId) {
         case REPAINT_TIMER_ID:
             win->delayedRepaintTimer = 0;
             KillTimer(hwnd, REPAINT_TIMER_ID);
-            win->RedrawAllIncludingNonClient(TRUE);
+            win->RedrawAllIncludingNonClient(true);
             break;
 
         case SMOOTHSCROLL_TIMER_ID:
--- a/src/Canvas.cpp
+++ b/src/Canvas.cpp
@@ -1182,12 +1182,24 @@ static LRESULT WndProcCanvasFixedPageUI(WindowInfo* win, HWND hwnd, UINT msg, WP
         case WM_GESTURE:
             return OnGesture(win, msg, wp, lp);
 
-        case WM_NCPAINT:
-            if (gGlobalPrefs->hideScrollbars) {
-                ShowScrollBar(win->hwndCanvas, SB_BOTH, FALSE);
+        case WM_NCPAINT: {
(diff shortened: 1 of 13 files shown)


src/EditAnnotations.cpp
```diff
@@ -113,7 +113,7 @@ AnnotationType gAnnotsWithColor[] = {
 // clang-format on
 
 // in SumatraPDF.cpp
-extern void WindowInfoRerender(WindowInfo*);
+extern void WindowInfoRerender(WindowInfo*, bool = FALSE);
```
kjk, line 116: When using bool, use true / false.
Also, name the bool argument. 
The rule is: if the argument is self-explanatory (like WindowInfo*), then we might

kjk, line 116: When using bool, use true / false.
Also, name the bool argument. 
The rule is: if the argument is self-explanatory (like WindowInfo*), then we might skip naming it. Otherwise, it should be named.

src/DisplayModel.h
```diff
@@ -135,6 +135,8 @@ struct DisplayModel : public Controller {
     void Relayout(float zoomVirtual, int rotation);
 
     Rect GetViewPort() const;
+    bool hScrollbarShown() const;
```
kjk, line 138: To be more in line with the rest of the code, this should be `IsHScrollbarVisible()`

src/Canvas.cpp
```diff
@@ -1312,7 +1330,7 @@ static void OnTimer(WindowInfo* win, HWND hwnd, WPARAM timerId) {
         case REPAINT_TIMER_ID:
             win->delayedRepaintTimer = 0;
             KillTimer(hwnd, REPAINT_TIMER_ID);
-            win->RedrawAll();
+            win->RedrawAllIncludingNonClient(TRUE);
```
kjk, line 1333: should be `true` because it's `bool`

src/SettingsStructs.h
```diff
@@ -44,6 +44,8 @@ struct FixedPageUI {
     Vec<COLORREF>* gradientColors;
     // if true, TextColor and BackgroundColor will be temporarily swapped
     bool invertColors;
+    // if true, hides the scrollbars but retain ability to scroll
+    bool hideScrollbars;
 };
```
kjk, line 49: See top of the file. The source of this is in `gen_settings_structs.go`. So add it there and then run `doit.bat -gen-structs` to regenerate this file. Requires go to be installed.

YongJieYongJie:
> 
> 
> Looks mostly good.
> 
> Could you also add context menu `Show Scrollbars" to menuDefContext (like "Show Bookmarks" etc.).
> See Menu.cpp and Commands.cpp.

Sure, let me look into that.

GitHubRulesOK:
> 
> 
> > Looks mostly good.
> > Could you also add context menu `Show Scrollbars" to menuDefContext (like "Show Bookmarks" etc.).
> > See Menu.cpp and Commands.cpp.
Is looking good for those wanting ScrollBarOff, is there a significant problem in having a similar ScrollBarOn for the vociferous majority that wish to have similarity to Acrobat Always On

kjk:
Thanks!

brontosaurusrex:
Is there a release planed with option to hide scrollbars?

Issue #1280, opened 2019-11-30
Add text search and translation from selection


kjk:
Still to do:
* [x] add top-level `Selection` menu after `Zoom`
* [x] more built-in services? Which ones?
* [x] a way for user to customize the list using `SelectionHandler` in advanced preferences, similar to `ExternalViewers`
* [ ] maybe: support invoking desktop apps. But which ones? Is it worth it?
* [ ] maybe: support [GoldenDic](http://goldendict.org/), it seems to accept a word on cmd-line

Desktop dictionaries:
* https://quest-app.appspot.com/home
* http://goldendict.org/
* http://www.tanos.co.uk/jlpt/extras/jwpce/

Other notes:
* https://github.com/Keypirinha/Packages/blob/master/WebSearch/websearch.ini : ideas for other websites

SumatraPeter:
Perhaps something along the lines of **ExternalViewers** in SumatraPDF-settings.txt, with appropriate levels of sanitization?

```
ExternalSites [
  [
    # URL of the website, may contain %s for selected text
    URL = https://www.google.com/search?q=%s

    # name of the website to be shown in the context menu (implied by domain name in URL if missing)
    Name =
  ]
]
```

GitHubRulesOK:
At present we can send external commands a filename [%1] and page number [%p] as in this example
![Annotator](https://user-images.githubusercontent.com/12754913/69907403-e2599800-13cc-11ea-8b9e-d747a81cd3ad.png)

In this case an image only page is sent to an "external editor" (MuPDF) where it can be modified and overwriten, this raises similar questions as to those such as when a music or video player application can be used to feed a recording application. Where should the boundary be ?

Current work in progress on selecting text could also allow "protected" word strings to be used in an arguably similar fashion ? Currently SumatraPDF only copies such text as an image ! however external apps can readily convert such strings to text and feed them to other commands such as a web search, text translator or TTS (text to speech). Adobe themselves have a sort of dual standard in allowing some restricted text to be used for screen readers.

GitHubRulesOK:
For a wscript vbs method see https://forum.sumatrapdfreader.org/t/advanced-tips-pass-selection-to-another-app-e-g-web-search/2700

Depends of course on you being allowed to run wscript and vbs files on your system

With the right parameters can be used for 
Add text search and translation from selection #1280
Search selection text in search engine #877
Search selected text in default internet browser #420
Add a translation feature #1351
[...]

PR #1757, opened 2020-09-10
#1280 add translation and searching of selected text

DhruvaSambrani:
# Introduction
As discussed on #1280, adding a feature to send selected text to external viewers with `%2`.

# Changes
1. Changed `ExternalViewers#ViewWithExternalViewer` to replace %2 with selected text
2. Changed `Menu#AppendExternalViewersToMenu` to add the index and shortcut to menu under `File` submenu
3. Changed `Menu#AddFileMenuItem` to remove the index of recent files

# ToDo
1. Update docs -> Mention the use of %2 and add some examples.
2. For multi-word selections, the url string must be enclosed in double quotes OR we need to implement URL safe encoding and first encode the selection before querying.

### DIFF
--- a/do/gen_settings_structs.go
+++ b/do/gen_settings_structs.go
@@ -389,7 +389,7 @@ var (
 	externalViewer = []*Field{
 		mkField("CommandLine", String, nil,
 			"command line with which to call the external viewer, may contain "+
-				"%p for page number and \"%1\" for the file name (add quotation "+
+				"%p for page number, %s for active selection and \"%1\" for the file name (add quotation "+
 				"marks around paths containing spaces)"),
 		mkField("Name", String, nil,
 			"name of the external viewer to be shown in the menu (implied by CommandLine if missing)"),
--- a/src/ExternalViewers.cpp
+++ b/src/ExternalViewers.cpp
@@ -348,20 +348,33 @@ bool ViewWithExternalViewer(TabInfo* tab, size_t idx) {
     }
 
     ExternalViewer* ev = gGlobalPrefs->externalViewers->at(idx);
+
     WStrVec args;
     ParseCmdLine(ev->commandLine, args, 2);
-    if (args.size() == 0 || !file::Exists(args.at(0))) {
+    if (args.size() == 0) {
         return false;
     }
 
+    bool isLink = false;
+    // Check if is URL, if not, check if path exists
+    if (str::StartsWith(args.at(0), L"http")) {
+        isLink = true;
+    } else if (!file::Exists(args.at(0))) {
+        return false;
+    }
     // if the command line contains %p, it's replaced with the current page number
     // if it contains %1, it's replaced with the file path
     // else if it contains %s it's replaced with active selection
     // else file path is appended
 
-    const WCHAR* cmdLine = args.size() > 1 ? args.at(1) : L"\"%1\"";
+    const WCHAR* cmdLine;
+    if (isLink) {
+        cmdLine = args.at(0);
+    } else {
+        cmdLine = args.size() > 1 ? args.at(1) : L"\"%1\"";
+    }
     AutoFreeWstr params;
-    bool toAppendFilePath = true;
+
     if (str::Find(cmdLine, L"%p")) {
         AutoFreeWstr pageNoStr(str::Format(L"%d", tab->ctrl ? tab->ctrl->CurrentPageNo() : 0));
         params.Set(str::Replace(cmdLine, L"%p", pageNoStr));
@@ -370,7 +383,6 @@ bool ViewWithExternalViewer(TabInfo* tab, size_t idx) {
     if (str::Find(cmdLine, L"%1")) {
         params.Set(str::Replace(cmdLine, L"%1", tab->filePath));
         cmdLine = params;
-        toAppendFilePath = false;
     }
     if (str::Find(cmdLine, L"%s")) {
         bool isTextOnlySelection;
@@ -380,13 +392,12 @@ bool ViewWithExternalViewer(TabInfo* tab, size_t idx) {
         }
         params.Set(str::Replace(cmdLine, L"%s", selection ? selection : L""));
         cmdLine = params;
-        toAppendFilePath = false;
     }
-    if (toAppendFilePath) {
-        params.Set(str::Format(L"%s \"%s\"", cmdLine, tab->filePath.Get()));
+    if (isLink) {
+        return LaunchBrowser(cmdLine);
+    } else {
+        return LaunchFile(args.at(0), params);
     }
-
-    return LaunchFile(args.at(0), params);
 }
 
 #define DEFINE_GUID_STATIC(name, l, w1, w2, b1, b2, b3, b4, b5, b6, b7, b8) \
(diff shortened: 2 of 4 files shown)


GitHubRulesOK:
I would not remove the "Open in" aspect but simply change to a more generic "Send &To"
ALT F# is a well established use so should not be altered e.g. ALT F1 should still open last file
Perhaps ensure that %2 has its own Top Menu area such as ALT &Query #

DhruvaSambrani:
@GitHubRulesOK So should I create a new submenu for External viewers? 
Where should I place it?
```
|
└-- File
└-- External Viewers
    └-- Open in Browser
    └-- Open in Adobe
```
This will result in a shortcut as `Alt-E #`
Or 
```
|
└--File
    └-- External Viewers
        └-- Open in Browser
        └-- Open in Adobe
````
This will result in a shortcut as `Alt-F E #`

**Comment by DhruvaSambrani** [comment]
@GitHubRulesOK Changed as requested
Query is External Viewers
So it goes as `Alt-E #`

**Comment by DhruvaSambrani** [comment]
@GitHubRulesOK @kjk Request to merge or suggest further changes.

**Comment by kjk** [comment]
This needs a list of at least 5 programs that can be used this way and how to launch them.
I assume you tested it with some software.
It's best if we auto-detect those programs the way we auto-detect AdobeReader etc.

**Comment by GitHubRulesOK** [comment]
@kjk 
I dont find the historic registry "auto detections" as useful as they may once have been, especially as in my case Acrobat, eXchange Foxit etc are all portable apps as are my Firefox Opera Chrome etc.

The stock search invocation of Internet explorer and Edge can be easily found at given locations using
"C:\Program Files\Internet Explorer\iexplore.exe" "%2"
"C:\Program Files (x86)\Microsoft\Edge\Application\msedge.exe" "%2"

What varies considerably is the user preference for a target of google, "bing translate", "websters dictionary" and those need to be set by the user.

Those could be added as defaults but many users have their own preferences so for PDF I currently use the common
\Apps\PDF\Adobe\Reader\AcroRd32.exe" /A page=%p "%1"
\Apps\PDF\FoxitReader\FoxitReader.exe /A page=%p "%1"
\Apps\PDF\PDFXEdit8\PDFXEdit.exe /A page=%p "%1"

For browser display thats
"C:\Program Files (x86)\Microsoft\Edge\Application\msedge.exe" file://"%1#page=%p"

but for searching bing 
"C:\Program Files (x86)\Microsoft\Edge\Application\msedge.exe" "_string with spaces_" would be very simple and for other apps %2 without quotes would also be great but I do agree it could be potentially dangerous without cheking first for a valid exe as at present.

**Comment by DhruvaSambrani** [comment]
I concur with @GitHubRulesOK about this. There are few PDF readers, but there are a lot of browser-SE pairings. A reasonable assumption is that everyone has Edge installed, and we can add "C:\Program Files (x86)\Microsoft\Edge\Application\msedge.exe" "%s" as a default. Any other websites or browsers should be added by the users. 
"Autodetection" can be done similar to that.

Also, there's a `WinUtils#launchBrowser(url)`. What is this? If it opens the default browser, then we can simply check if the CommandLine argument is a url, and call `WinUtils#launchBrowser(url_with_replacements)`.

**Comment by DhruvaSambrani** [comment]
So in the most recent commit, I've added support to open URLs in the default browser. If CommandLine starts with `http` then we assume that it is a link, and instead of calling `LaunchFile`, we call `LaunchBrowser`. This way, users can simply add 
```
[
    CommandLine = "https://duckduckgo.com?q=%s"
]
```
Which adds a "Send to duckduckgo" in External Viewers, and opens it in the default browser.
They can also add the usual form of `CommandLine = path_to_exe url_with_%2` for other browsers as installed

**Comment by DhruvaSambrani** [comment]
@kjk @GitHubRulesOK I'm relatively free over the weekend. Let me know if any changes are necessary.

**Comment by GitHubRulesOK** [comment]
As an "observer" I cant compile (rusty skills and poor hardware, make me a "leacher") If you have a working copy to hand I can make suggestions but it is up to @kjk whether it passes his aproval for PR (public release:-)

**Comment by kjk** [comment]
@DhruvaSambrani at this point it's not about the code. When people use Sumatra, it should come with built-in "search on Google", "translate to French", "translate to German" option (i.e. the most common) + a link to a doc that says how to configure more options.

In other words, I need a list of as many websites / windows programs that can be sub-launched this way so that I can provide the most common options built-in and then make it easy for users to configure more.

**Comment by DhruvaSambrani** [comment]
## Search engines - 
Ecosia - https://www.ecosia.org/search?tt=a1b7da50&q=%s
Wikipedia - https://en.wikipedia.org/wiki/Special:Search?search=%s
Google - https://www.google.com/search?q=%s
Duckduckgo - https://duckduckgo.com/?q=%s

## Translators
Google Translate - https://translate.google.com#view=home&op=translate&sl=`fromLanguageCode`&tl=`toLanguageCode`&text=%s
Bing translate - https://www.bing.com/translator?from=`fromLanguageCode`&to=`toLanguageCode`&text=%s

## Dictionaries
Webster - https://www.merriam-webster.com/dictionary/%s
Oxford - https://www.oxfordlearnersdictionaries.com/definition/english/%s
Cambridge - https://dictionary.cambridge.org/dictionary/english/%s

However I still feel that there shouldn't be any default urls added, because of the highly personalized nature of this set up. Your example of French and German is not going to be used by a lot of people, for whom those options are simple bloat. Same with Search Engines and dictionaries.

👍  for the link to doc.

**Comment by DhruvaSambrani** [comment]
@kjk

**Comment by DhruvaSambrani** [comment]
@kjk @GitHubRulesOK Bump.

**Comment by GitHubRulesOK** [comment]
I am all for keeping it simple, it does not matter if Edge, Firefox or Netscape is my choice of browser, if I have selected an unknown peice of foreign, to me text, I will expect to see it in my current browser language.

For example if the string is = `Dia duit an Domhan`
I can expect the default browser in my language will call a named translator, so just SendTo windows as a command cmd.exe /c  or in SumatraPDF terms Open in ExternalViewers 
```
start "Translating" "https://translate.google.com/?&text=Dia duit an Domhan"
start "Translating" "https://translate.google.com/?&text=こんにちは世界"
```

DhruvaSambrani:
Ah ok, I probably misunderstood
So is this ready to merge?

GitHubRulesOK:
No misunderstanding it is @kjk who decides what is acceptable to him.
My simplistic request was to pass a string of any language to a users choice of action whether that be an offline grammer/dictionary app or a websearch. kjk wants it to be a simple set of default choices that are well defined and documented, such as to reduce new user problems.
My experience is that in this case there may be a few common desires (Web translate/Search/Define Word) but eventually the requests will be how do I change the defaults. However my simplistic approach could potentially lead to abuse.

DhruvaSambrani:
@kjk Bump

DhruvaSambrani:
@kjk Bump

DhruvaSambrani:
@kjk Bump

DhruvaSambrani:
@kjk bump

kjk:
@DhruvaSambrani Sorry for sitting on this for so long.

I've started on this functionality in https://github.com/sumatrapdfreader/sumatrapdf/commit/d710475e28acc407cb7fa6fc08e00d77ef8753a9

Still more to do, you can track progress / add ideas in https://github.com/sumatrapdfreader/sumatrapdf/issues/1280

PR #1776, opened 2020-10-11
Make number of most recently used files configurable

pawelszramowski:
I do not find 10 recently used files to be nearly enough so I've made the number configurable through SumatraPDF-settings.txt.

### DIFF
--- a/do/gen_settings_structs.go
+++ b/do/gen_settings_structs.go
@@ -626,6 +626,10 @@ var (
 			"if true, we show a list of frequently read documents when no document is loaded"),
 		mkField("UseTabs", Bool, true,
 			"if true, documents are opened in tabs instead of new windows").setVersion("3.0"),
+		mkField("FileHistoryMaxRecent", Int, 10,
+			"number of most recently used files that will be shown in the menu "+
+			"(and remembered in the preferences file, if just filenames are "+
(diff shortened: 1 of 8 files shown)


GitHubRulesOK:
I am not a C++ programmer but the removal of a limit suggests the value is unbounded?
 what happens if I set it to 10000000 or -1 or "false"?

pawelszramowski:
Non-integer values are coerced to 0, negative are ignored, positive are used. You'd have to have opened 10000000 files before such a limit kicks in.

pawelszramowski:
@kjk are you going to merge it? Would you like me to introduce some additional changes?

kjk:
Sorry for the late reply. 

Until we can scroll the thumbnails, I don't want to introduce configurability. Increasing the number of thumbnails without ability to scroll them is not very useful.

Adding scrolling is probably a pretty big task.

pawelszramowski:
But wait! It has nothing to do with thumbnails, shown when there are no open files, a number of which is controlled by `#define FILE_HISTORY_MAX_FREQUENT`. It only makes configurable a number of most recently used files, shown in the File menu, previously controlled by `#define FILE_HISTORY_MAX_RECENT`.

pawelszramowski:
@kjk not sure if you've noticed my previous comment so calling you out again, maybe you'll reconsider...

Issue #723, opened 2017-02-15
Buttons for Page Rotate


jitenderdogra:
Please add buttons for page rotate clockwise and counterclockwise beside the zoom in and zoom out buttons. It is cumbersome to open the menu or having to press three buttons each time i want to rotate the page.

Thank you.

ChesterCloud:
I agree with you. This function should be there.

gschenck:
You can use the shortcut StrgShift+ for rotating clockwise or StrgShift- for rotating other direction.

Petewg:
Indeed, that's an highly needed feature, especially for landscaped pages.
Maybe, it'd be even better adding an `auto-rotate` advanced setting 
(f.e.: AutoRotateLandscapePages = right )
Of course, such an option would require sumatraPDF  be able to recognize page orientation. 

PS. (like every more-than-two-keys-combination), current key combination for rotate `Ctrl+Shift++`  is quite unhandy/complex-to-remember.

PS.2. no matter what, SumatraPDF is great!

Petewg:
FYI.
I've found undocumented (?) shortcuts for page rotation that are quite simpler than complex `Ctrl+Shift+<+/->`  combination.
 - Numeric keypad: `*` (VK_MULTIPLY) --> clockwise (90) rotation 
- Numeric keypad:  `/`  (VK_DIVIDE) --> counterclockwise (-90) rotation.

mutch easier to use/remember; perhaps it should be added in  [manual](https://www.sumatrapdfreader.org/manual.html)?

Massedil:
It could also be great to have the ability to rotate the document in the right clic menu.

GitHubRulesOK:
**The shortest buttons for rotate are** 
Anti-clockwise=<kbd>**[**</kbd> 
and clockwise <kbd>**]**</kbd>

kjk:
Added rotate buttons to toolbar in https://github.com/sumatrapdfreader/sumatrapdf/commit/72d9a9418e224362461ddf7598ed8601fbcd2aa2

GitHubRulesOK:
Just for fun I tried this with a comic book, whilst screen was rotated it works but @ 90 degrees mouse movement is awkward, (may work better for touchscreen users)
![image](https://user-images.githubusercontent.com/12754913/107294500-6b566f80-6a65-11eb-9683-f14c2d80e7d7.png)

Petewg:
> 
> 
[...]

PR #1875, opened 2021-01-31
add rotate buttons to the toolbar

pawelzwronek:


GitHubRulesOK:
@kjk 
very very popular long time request (there have even been forks just include this feature as its frequently wanted by some users with technical / financial documents)

@pawelzwronek 
very minor spelling mistake und of comments line 132 (should be "modified")

pawelzwronek:
Fixed. Yeah, you can close #723.
Do you like the position of buttons?

src/Toolbar.cpp
```diff
@@ -69,8 +69,8 @@ static ToolbarButtonInfo gToolbarButtons[] = {
     {5, CmdZoomFitPageAndSinglePage, _TRN("Fit a Single Page"), 0},
     {6, CmdZoomOut, _TRN("Zoom Out"), 0},
     {7, CmdZoomIn, _TRN("Zoom In"), 0},
-    {13, CmdViewRotateLeft, _TRN("Rotate Left"), 0},
-    {14, CmdViewRotateRight, _TRN("Rotate Right"), 0},
+    {13, CmdViewRotateLeft, _TRN("Rotate &Left\tCtrl+Shift+-"), 0},
```
GitHubRulesOK, line 72: @kjk can correct me 
but I do not think this is the correct place for & shortcuts since they are provided elsewhere by _TRN

pawelzwronek: I'm not sure, but without a full text I copied from `strings\translations.txt` the translation is not working 🤔

GitHubRulesOK: I have to admit without comprehending  C++
 I do not understand all the ways the sections interact, but since translation works for other icons, I see no reason the extra ones do not follow UNLESS there is a limiting zone of action (but these are included amongst others) or index number limit (perhaps a 10 needs to be 12 ?)

pawelzwronek:
Other icons have no key shortcuts. I think is was much easier to include
key combination straight to a command name than creating second label next
to it.

niedz., 31 sty 2021, 14:21 użytkownik GitHubRulesOK <
notifications@github.com> napisał:

GitHubRulesOK:
The other icons should all have keyboard shortcuts (not always listed in menu) but their translations are functioning as tooltips without a need to show their controls. so French translation is visibly different but + & - keys work on QWERTY keyboard same as the + & - icons
Is it perhaps language specific? did you try switching to another language you recognise ? but note there is also [ & ] for rotation in addition to + & _ (note it is underline when shifted)

src/Toolbar.cpp
pawelzwronek: Other icons have no key shortcuts. I think is was much easier to include
key combination straight to a command name than creating second label next
to it.

pawelzwronek: Other icons have no key shortcuts. I think is was much easier to include
key combination straight to a command name than creating second label next
to it.

pawelzwronek: GitHubRulesOK: "The other icons should all have keyboard shortcuts (not always listed in menu) but their translations are functioning as tooltips without a need to show their controls. so French translation is visibly different but + & - keys work on QWERTY keyboard same as the + & - icons
Is it perhaps language specific? did you try switching to another language you recognise ? but note there is also [ & ] for rotation in addition to + & _ (note it is underline when shifted)"

pawelzwronek: I see some redudancy e.g. in a translation of "Copy Selection" in `strings\translations.txt`. Once it's used in menu with a key shorctuct, next it's used in context menu(RMB) but without the key shortcut. Both have separate translaton in `strings\translations.txt` but it's basicly the same command. If I should remove the shortcut I have to create new translation without it.
Is see no keyboard shortucts on tooltips.

pawelzwronek: Sorry for the mess. I replied by gmail 😞

kjk:
Thanks, I've made https://github.com/sumatrapdfreader/sumatrapdf/commit/72d9a9418e224362461ddf7598ed8601fbcd2aa2 based on this PR (without renaming of the toolbar.bmp)

Issue #1891, opened 2021-02-19
Zoom in value jumps between two values when max zoom is reached


WhiteChairFromIkea:
Steps to reproduce:
1. Open PDF;
2. CTRL + Mouse wheel up, or click zoom in icon many times;
3. Zooming level begins to jump between "max" to "max -1" after maximum zoom level is reached;

![image](https://user-images.githubusercontent.com/46539653/108498280-51166000-72b5-11eb-83e3-09e8ede4bbfc.png)

https://user-images.githubusercontent.com/46539653/108498330-63909980-72b5-11eb-8cc4-a305b02bc8d6.mp4

GitHubRulesOK:
@kjk 
This odd behaviour of alternating from max value (6400) to drop back to max custom/default lower (4800) has been there for some time so is most noticeable in default settings.
As a **workaround** if wishing to reduce visual effect it can be significantly reduced by adding 6399 as a custom value to fall-back to, but then seems to be affected by "pixel creep" (down and right) of some values,

kjk:
I can't reproduce this: I've tested 3 PDF files and I just reach 6400 and stay there.

I'm reluctant to add fixes if I can't verify the fix.

Does it happen to you with every PDF file in every view?

Could you provide a sample PDF and more repro info (like: view used etc.).

kjk:
I was only testing on max zoom, which didn't reproduce, but it did reproduce on min zoom.

GitHubRulesOK:
Fixes my **Max** zoom problem (i had not noticed on Min, but had not been looking there closely :-)

WhiteChairFromIkea:
Hi, neither MAX (happened on all pdfs in my setup) nor MIN (never experiences) jumping does not occur anymore in:
![image](https://user-images.githubusercontent.com/46539653/108667548-23fac500-74e2-11eb-814b-5c1cc54ba4d6.png)

Tested twice

PR #1892, opened 2021-02-19
Fix zoom value jump when at ZOOM_MAX or ZOOM_MIN

aleksandar-j:
Addresses issue #1891. When within FUZZ distance of limit values, don't zoom.

I am not sure if this is the optimal solution to this problem and whether constant FUZZ should be used in this manner but zoom still works and the annoying problem is gone.

### DIFF
--- a/src/DisplayModel.cpp
+++ b/src/DisplayModel.cpp
@@ -1591,7 +1591,7 @@ float DisplayModel::GetNextZoomStep(float towardsLevel) const {
 
     const float FUZZ = 0.01f;
     float newZoom = towardsLevel;
-    if (currZoom < towardsLevel) {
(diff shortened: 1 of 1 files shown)


kjk:
Thanks. I've used this fix in https://github.com/sumatrapdfreader/sumatrapdf/commit/4146551a6cc85b53a8721a7c522c18b9b6554099

aleksandar-j:
@kjk 
Any particular reason my PR got closed? I could understand if you wanted to edit something and didn't want to initiate a back and forth with me for such a trivial change; but as far as I can tell, your commit is just my fix verbatim + unrelated refactoring...

Regardless, great software and I am glad this thing finally got fixed as this has been bothering me for quite some time.

PR #1940, opened 2021-05-05
Corrected broken home url of various files

tasoss:
Corrected broken url of settings.html


GitHubRulesOK:
using search I find there is also similar in "Customize-external-viewers.html"

tasoss:
thanks @GitHubRulesOK  i will fix it too.

GitHubRulesOK:
that's great however for bonus points I did wider search and got another 5 x http://www.sumatrapdfreader.org hits in the website area
https://github.com/search?q=%22http%3A%2F%2Fwww.sumatrapdfreader.org%22++repo%3Asumatrapdfreader%2Fsumatrapdf+extension%3Ahtml+path%3A%2Fwebsite&type=code

tasoss:
Sigh...
I think it should be ok now.
I have changed it also in notion_cache dir. 
Should notion_cache dir be included here?

kjk:
Thanks, but the source of this is https://www.notion.so/SumatraPDF-documentation-fed36a5624d443fe9f7be0e410ecd715 and that's where it needs to be fixed. The .html and notion_cache files are auto-generated with a script.

I think I fixed links in Notion and re-generated the files.

PR #1942, opened 2021-05-09
Added close all files option

itsrdb:
Added 'Close All Files' option under 'File->Close All Files' to close all the tabs simultaneously

### DIFF
--- a/src/Commands.h
+++ b/src/Commands.h
@@ -10,7 +10,7 @@ Cmd* enum (e.g. CmdOpen) and a human-readable name (not used yet).
     V(CmdOpen, "Open File...")                                            \
     V(CmdOpenFolder, "Open Folder...")                                    \
     V(CmdClose, "Close Document")                                         \
-    V(CmdCloseAll, "Close Document")                                      \
+    V(CmdCloseAll, "Close All Documents")                                 \
     V(CmdSaveAs, "Save File As...")                                       \
     V(CmdPrint, "Print Document...")                                      \
     V(CmdShowInFolder, "Show File In Folder...")                          \
--- a/src/Commands.h
+++ b/src/Commands.h
@@ -10,6 +10,7 @@ Cmd* enum (e.g. CmdOpen) and a human-readable name (not used yet).
     V(CmdOpen, "Open File...")                                            \
     V(CmdOpenFolder, "Open Folder...")                                    \
     V(CmdClose, "Close Document")                                         \
+    V(CmdCloseAll, "Close Document")                                      \
     V(CmdSaveAs, "Save File As...")                                       \
     V(CmdPrint, "Print Document...")                                      \
     V(CmdShowInFolder, "Show File In Folder...")                          \
--- a/src/Menu.cpp
+++ b/src/Menu.cpp
@@ -94,7 +94,7 @@ static MenuDef menuDefFile[] = {
     { _TRN("&Open...\tCtrl+O"),             CmdOpen,                   MF_REQ_DISK_ACCESS },
     { "Open Folder",                        CmdOpenFolder,             MF_REQ_DISK_ACCESS | MF_RAMICRO_ONLY },
     { _TRN("&Close\tCtrl+W"),               CmdClose,                  MF_REQ_DISK_ACCESS },
-    { _TRN("&Close All Files\tCtrl+Shift+W"),  CmdCloseAll,             MF_REQ_DISK_ACCESS },
+    { _TRN("Close All &Tabs\tCtrl+Shift+W"),  CmdCloseAll,             MF_REQ_DISK_ACCESS },
     { _TRN("Show in &folder"),              CmdShowInFolder,           MF_REQ_DISK_ACCESS },
     { _TRN("&Save As...\tCtrl+S"),          CmdSaveAs,                 MF_REQ_DISK_ACCESS },
     { _TRN("Save Annotations"),             CmdSaveAnnotations,        MF_REQ_DISK_ACCESS },
--- a/src/Menu.cpp
+++ b/src/Menu.cpp
@@ -94,6 +94,7 @@ static MenuDef menuDefFile[] = {
     { _TRN("&Open...\tCtrl+O"),             CmdOpen,                   MF_REQ_DISK_ACCESS },
     { "Open Folder",                        CmdOpenFolder,             MF_REQ_DISK_ACCESS | MF_RAMICRO_ONLY },
     { _TRN("&Close\tCtrl+W"),               CmdClose,                  MF_REQ_DISK_ACCESS },
+    { _TRN("&Close All Files"),             CmdCloseAll,               MF_REQ_DISK_ACCESS },
     { _TRN("Show in &folder"),              CmdShowInFolder,           MF_REQ_DISK_ACCESS },
     { _TRN("&Save As...\tCtrl+S"),          CmdSaveAs,                 MF_REQ_DISK_ACCESS },
     { _TRN("Save Annotations"),             CmdSaveAnnotations,        MF_REQ_DISK_ACCESS },
@@ -574,6 +575,7 @@ static void MenuUpdateStateForWindow(WindowInfo* win) {
     // TODO: happens with UseTabs = false with .pdf files
     SubmitCrashIf(IsFileCloseMenuEnabled() == win->IsAboutWindow());
     win::menu::SetEnabled(win->menu, CmdClose, IsFileCloseMenuEnabled());
+    win::menu::SetEnabled(win->menu, CmdCloseAll, IsFileCloseMenuEnabled());
 
     MenuUpdatePrintItem(win, win->menu);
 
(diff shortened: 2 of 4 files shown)


itsrdb:
Hi! @kjk Have a look at this feature I've added and also suggest if there's any improvement that needs to be done.

src/Commands.h
```diff
@@ -10,6 +10,7 @@ Cmd* enum (e.g. CmdOpen) and a human-readable name (not used yet).
     V(CmdOpen, "Open File...")                                            \
     V(CmdOpenFolder, "Open Folder...")                                    \
     V(CmdClose, "Close Document")                                         \
+    V(CmdCloseAll, "Close Document")                                      \
```
GitHubRulesOK, line 13: "Close Documents")

src/Menu.cpp
```diff
@@ -94,6 +94,7 @@ static MenuDef menuDefFile[] = {
     { _TRN("&Open...\tCtrl+O"),             CmdOpen,                   MF_REQ_DISK_ACCESS },
     { "Open Folder",                        CmdOpenFolder,             MF_REQ_DISK_ACCESS | MF_RAMICRO_ONLY },
     { _TRN("&Close\tCtrl+W"),               CmdClose,                  MF_REQ_DISK_ACCESS },
+    { _TRN("&Close All Files"),             CmdCloseAll,               MF_REQ_DISK_ACCESS },
```
GitHubRulesOK, line 97: Consider if could be ALT+F+X  / CTRL + X 
Thus reduce to shorter "E&xit All Tabs\tCtrl+X"),

itsrdb: CTRL + X could be accidently triggered while copying some text, I believe that ALT + F + X would be a better choice

GitHubRulesOK: There is another conflict that needs to be clarified since lower
{ _TRN("E&xit\tCtrl+Q"),                CmdExit, 

Thus X is taken and agree CTRL X is common usage
perhaps  a different combo is needed

itsrdb: Maybe some function key + CTRL/ALT? :( This is hard to think

itsrdb: I gave a thought about adding a shortcut earlier but realized that it shouldn't be triggered accidently nor should the shortcut be difficult to remember. Sublime Text didn't have a shortcut for closing all files but Notepad++ has CTRL + SHIFT + W

itsrdb: So I have added CTRL+SHIFT+W, Should I commit these changes or implement some other shortcut keys?

GitHubRulesOK: Ctrl W is close windowed tab
 so Ctrl SHIFT close Windowed tabs  sounds good for accelerated exit
just need to decide on the Alt F+ wording

GitHubRulesOK: Perhaps Alt F T but need to double check
 Close All &Tabs

itsrdb: Should Alt F T work the same or the two shortcuts are different?

GitHubRulesOK: so may need Accelerators.cpp Line after https://github.com/sumatrapdfreader/sumatrapdf/blob/802852a648f05eb5bff360dd3a0d01b03da4e0a0/src/Accelerators.cpp#L28

 {FSHIFT | FCONTROL | FVIRTKEY, 'W', CmdCloseAll},

itsrdb: Yes, it's done. I thought you wanted to me add two shortcuts CTRL+SHIFT+W and ALT+F+T

GitHubRulesOK: ALT F + values are dynamic so &Acrobat &Foxit &Microsoft etc
but can also be different for each language, thus get comments and frequent changes
also there are other examples of non matching equality such as the e&xit = Ctrl Q 
the aim is to minimise "English" confusion by using key words such as Tabs

itsrdb: I think I would need your help on this one, I'm making the commit for CTRL+SHIFT+W shortcut for now

GitHubRulesOK: I dug up my notes for those I use for addins

# Beware many ALT+F+& shortcut combinations are already used thus adding a 2nd
# can add to confusion especially when they may conflict with non-English ones.

# Default Engish Keys = W O C F(x2) S [H] N P A F(x2) I M(MutualyExc.x2) E R X

So I should have known X was already taken

# So MY extras are ALT+F+ M(2) / M(2) / D / L / U / G / Q "/" & #
However **as unofficial** they can be used in Core application

I may have also used some of the following including T but as above these are SumatraPDF+ uses
# Free= B J K T V Y Z  we can also use other basic symbol keys ` = [ ] ; ' \ ,
# Unless on the keypad * - + . Shifted are more dificult ¬!"£$%^&()_{}:@~|<>? 

SO IN SUMMARY it looks like <kbd>ALT</kbd> <kbd>F</kbd> <kbd>T</kbd> is a good candidate

itsrdb: Can you suggest the changes to be made and where? lol I really am confused with these shortcuts sorry, but ALT F T is final now

src/Menu.cpp
```diff
@@ -94,7 +94,7 @@ static MenuDef menuDefFile[] = {
     { _TRN("&Open...\tCtrl+O"),             CmdOpen,                   MF_REQ_DISK_ACCESS },
     { "Open Folder",                        CmdOpenFolder,             MF_REQ_DISK_ACCESS | MF_RAMICRO_ONLY },
     { _TRN("&Close\tCtrl+W"),               CmdClose,                  MF_REQ_DISK_ACCESS },
-    { _TRN("&Close All Files"),             CmdCloseAll,               MF_REQ_DISK_ACCESS },
+    { _TRN("&Close All Files\tCtrl+Shift+W"),  CmdCloseAll,             MF_REQ_DISK_ACCESS },
```
GitHubRulesOK, line 97: &Close is a conflict so am suggesting
Close All &Tabs

itsrdb: It's done 👍

src/Menu.cpp
```diff
@@ -94,7 +94,7 @@ static MenuDef menuDefFile[] = {
     { _TRN("&Open...\tCtrl+O"),             CmdOpen,                   MF_REQ_DISK_ACCESS },
     { "Open Folder",                        CmdOpenFolder,             MF_REQ_DISK_ACCESS | MF_RAMICRO_ONLY },
     { _TRN("&Close\tCtrl+W"),               CmdClose,                  MF_REQ_DISK_ACCESS },
```
GitHubRulesOK, line 96: @kjk 
NOT REQUIRED since it would need an extra translation changed entry ?
BUT existing &Close could be expanded to &Close Tab
Thus showing singularity distinct from the new multiple

GitHubRulesOK:
@kjk 
I am unclear as to how this affects non tabbed usage OR TWO instances
a)in terms of wording
b) functionality e.g. is an extra test required for 2 windows and flushing session data on trigger

src/Menu.cpp
itsrdb: I'm sorry I'm unaware of how this translation feature works so any suggestions are appreciated, should I then change the originally written '&Close' to 'Close &Tab' as suggested?

itsrdb: Also the ALT F T shortcut is yet to be written, does it have to be written in accelerators.cpp?

GitHubRulesOK: @itsrdb
kjk will usually deal with translation as a separate requirement thus am NOT insisting existing is changed as there will be wider impact on existing translations, especially since it is used in a wider context.
the alt ft is automatic by adding the &Tab

GitHubRulesOK: P.S If all is well you may wish to follow up after with website additions such as 
1 or 2 "Manual" pages
but wait till @kjk review is completed

itsrdb: Sure @GitHubRulesOK

kjk:
Sorry but one of the guiding principles of Sumatra is simplicity. 

Adding more menu items adds complexity to the UI so there must be a good reason to add it i.e. we only add menus for frequently used functionality.

The code looks fine  but I don't believe this is frequently added functionality.

GitHubRulesOK:
@kjk it is common functionality for many tabbed applications, and is frequently raised as an issue or comment on forum, one possible reason there are not more requests is some users see it was requested almost as soon as someone opened 20 tabs
image

kjk:
I could see it as a "silent" option under Ctrl-Shift-W (or Ctrl-Alt-W) shortcut or when we have context menu on tabs, then it could be part of that context menu.

itsrdb:
@kjk Many people have asked for this feature in the past including me and this was planned while keeping other text/pdf readers in mind https://forum.sumatrapdfreader.org/t/how-to-close-all-tabs-with-one-click/1102/4 The dynamic shape change of tab bars on pressing the close button makes it hard to close all the tabs consecutively

GitHubRulesOK:
@kjk for minimal "Silent" menu space  we could for now just add to manual not menu 
Close Tab (Shift for All)   Ctrl+W
However I personally think ALL BUT is preferable

Issue #1895, opened 2021-02-26
[feature] to collapse all bookmarks the same level


Aliaksandr3by:
[Option] to collapse all bookmarks of the same level

![image](https://user-images.githubusercontent.com/11432724/109243320-c6ba8880-77ed-11eb-92e6-ed9b8e9747aa.png)

GitHubRulesOK:
I think this is a duplicate but will take some time to search the active one

Aliaksandr3by:
I searched it, but do not found:
I asked that function on the official site .

kjk:
Is there other software having such functionality so that I can see how they implement it?

Instead of adding 2 more context menu items, could make expand / collapse dual purpose:
* when clicked on a tree node with children, expands / collapses recursively only this node
* when clicked elsewhere, expands // collapses everything

PR #1969, opened 2021-06-05
#1895 Added bookmark option "Expand all within selected level" and "Collapse all within selected level"

Ashfaaq18:
Added this option as discussed in issue #1895 and the [forum post](https://forum.sumatrapdfreader.org/t/add-ability-to-collapse-all-bookmark-elements-at-the-same-level-and-their-descendants/3237).

Collapse all within selected level:
collapseAllWithinSelectedLevel

Expand all within selected level:
expandAllWithinSelectedLevel

### DIFF
--- a/mupdf/include/mupdf/pdf/annot.h
+++ b/mupdf/include/mupdf/pdf/annot.h
@@ -235,9 +235,8 @@ fz_link *pdf_create_link(fz_context *ctx, pdf_page *page, fz_rect bbox, const ch
 	create a new annotation of the specified type on the
 	specified page. Populate it with sensible defaults per the type.
 
-	The page takes a reference, and an additional reference is
-	returned to the caller, hence the caller should drop the
-	reference once it is done.
+	The returned pdf_annot structure is owned by the page and does
+	not need to be freed.
 */
 pdf_annot *pdf_create_annot(fz_context *ctx, pdf_page *page, enum pdf_annot_type type);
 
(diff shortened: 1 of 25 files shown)


GitHubRulesOK:
Not the developer so its down to @kjk 

It is a reasonable addition to add control for grand-children but will only be seen as adding to the issues since both open requests are very clear that their "desire" is to control `siblings@level` 

So although they have not clearly considered the consequences both their requirement in your first illustration is to click on 7 and 8 will collapse ! since they can easily collapse all of 7 themselves

@ Level means pick any level and AND ALL at that level be expanded or contracted.
However I think you will then find they say that is not exactly what they wanted,
I am certain the adage "You can't please ALL ... = true

Ashfaaq18:
I just realized my misunderstanding, Do you mean like this? 

image

the option should collapse or expand all level 2 siblings if level 2 is selected
the option should collapse or expand all level 1 siblings if level 1 is selected

correct me if I'm wrong.

GitHubRulesOK:
That is my understanding, the term "levels" is usually a reference to the amount of "nesting" AKA "indented" as shown by 1.0 & 2. & 3 is counted as one level and 1.10 2.1 3.999 is referenced as the "next" level

kjk:
Thanks. Have to think about it, maybe after 3.3 is released.

Ashfaaq18:
Sure, ill get back to this when you want it implemented. Do let me know if I've got to change it according to my 2nd comment or leave it as it is for the time being.

PR #2004, opened 2021-06-25
when printing, detect printer paper size when it's not set explicitly

M-HT:
When printing and when the paper size is not set explicitly, this code searches the list of paper sizes supported by the printer and if a paper size matches the size of first page in the document, then this paper size is used.


kjk:
Thanks!

PR #2009, opened 2021-06-29
when printing, use custom paper size when it's not set explicitly / detected

M-HT:
When printing and when the paper size is not set explicitly, nor is it detected in the list of paper sizes supported by the printer, this code sets a custom paper size based on the first page in the document.


kjk:
Thanks!

Issue #2024, opened 2021-07-12
Regression: With multiple windows open, SessionState is not properly saved on application exit


hg421:
## Issue
In version 3.3 there is a bug with saving the session state on application exit (by File->Exit, or <kbd>Ctrl</kbd>+<kbd>Q</kbd>) with more than one window open:

In this case, the state of **only one** window is being stored in the `SessionState` array, while the states of the other windows are lost.

With version 3.2 all windows were saved (and hence, restored on next launch).

## Cause
This problem looks to be an unfortunate side effect of the fix for issue #1659 (which was that the session state was only saved on exit, hence was lost if Sumatra didn't terminate gracefully).
Now, the session state is always kept up to date over the application lifetime.
Specifically, commit 351f73d1e2b8269b3a6262987e72c3ea21727fd0 changes `prefs::Save()` so that it always calls `RememberSessionState`, which means that on each call to `prefs::Save()` we store the current state of all windows to disk.
In particular, this happens every time a window is closed by `CloseWindow`.

Now, the problem is that `OnMenuExit` closes all the windows by iterating over them and calling `CloseWindow` for each one:
https://github.com/sumatrapdfreader/sumatrapdf/blob/8578d1723a7beac5eb627e3cc0dac2169a92fad1/src/SumatraPDF.cpp#L2157-L2160
Consequently the current session state (when some windows may be already closed) is getting saved before **each window closing**. And what finally ends up in the settings file is the state before closing the *last* window, which explains that only one window remains.

By the way, in version 3.2 this problem was avoided because `RememberSessionState` used to be called only *once*  by `OnMenuExit` but then not by `CloseWindow`. See old code in `CloseWindow`:
https://github.com/sumatrapdfreader/sumatrapdf/blob/ff1e03f71c9783273ad1ec2562d6265ddc091a87/src/SumatraPDF.cpp#L2390-L2395
(note: since, before the fix of #1659, sessionData was normally empty when the app was running, it could check that to find out whether `OnMenuExit` had saved it already...)

## Fix
I'm not sure what is the best way to fix this... we would probably have to prevent `RememberSessionState` from being called while `OnMenuExit` is closing the windows one by one...
(e.g. separating `RememberSessionState` from `prefs::Save()` again; or have some way to tell `prefs::Save()` not to call `RememberSessionState`?)

GitHubRulesOK:
I would need to look more closely. The initial problem with 3.1.2 was NO session data held until clean closure.
[...]

PR #2027, opened 2021-07-13
workaround for #2024: don't save partial app state during termination

hg421:
See issue #2024.
Adds a global flag `gDontSavePrefs` so that on app exit the session state is saved by `OnMenuExit` and not by `CloseWindow`.

Notes:
- if `gDontSavePrefs` is set, all prefs are prevented from being saved by `CloseWindow` (also those apart from just the session state).  But it's probably fine as it looks like there are no relevant setting updates after we start to close the windows...
- `CloseWindow` tries to save the window position/size with `RememberDefaultWindowPosition` which won't get picked up if the flag is set, but actually `RememberSessionState` will have already done that for all open windows so that's ok. And by my local tests the window sizes are in fact correctly remembered when launched the next time.

Feedback welcome!

### DIFF
--- a/src/SumatraPDF.cpp
+++ b/src/SumatraPDF.cpp
@@ -174,6 +174,10 @@ static WStrVec gAllowedLinkProtocols;
 // on an in-document link); examples: "audio", "video", ...
 static WStrVec gAllowedFileTypes;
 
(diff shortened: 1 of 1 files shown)


kjk:
looks good to me!

Issue #2001, opened 2021-06-21
Save state before ctrl-shift-n


kjk:
This is so that the file opens in a new window at the same position.

Maybe add ctrl-alt-n for opening in a new tab.

Related to #46

GitHubRulesOK:
I will close https://github.com/sumatrapdfreader/sumatrapdf/issues/1448# as a more muddled older duplicate of this description. For illustration, here is a recent screenshot where the file was last stored as closed/opened at page 15 but a new window on page 91 will thus open the second copy at page 15.
https://github.com/sumatrapdfreader/sumatrapdf/issues/1448#

PR #2032, opened 2021-07-16
Honoring links destinations zoom level

Lucas-C:
I'd like to contribute to Sumatra PDF so that it honors the zoom level of "XYZ" internal links
(cf. section 12.3.2.2 Explicit Destinations of the [PDF 1.7 spec](https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf)).
Adobe Acrobat Reader takes it into consideration, and I'd like Sumatra PDF to also do so
(in order for me to build game mechanics on it for a new PDF game).

My questions so far:
* [x] are you OK to welcome this feature in Sumatra PDF?
* [x] is it OK to add some unit tests in `src/utils/tests/EngineFzUtil_ut.cpp` in order to test a function of the `src/utils/` directory? And if not, where could I put those tests?

What remains to be done (at least):
* [x] fix the code added to `DisplayModel::ScrollToLink` in order to properly scale & set the new zoom level
* [x] ensure that after following a `/XYZ` link with a zoom > 0 Sumatra view port is positioned at the correct position (x, y, zoom) compared to Adobe Reader behaviour


kjk:
> are you OK to welcome this feature in Sumatra PDF?
Yes.

> unit tests in src/utils/tests/EngineFzUtil_ut.cpp 

You can add tests to `SumatraUnitTests.cpp` or `src/EngineFzUtil_ut.cpp`. If we had more unit test files for code in `src`, then we could create `src/tests`

kjk:
the way to to modify vs project / solution files is to edit `premake5.files.lua` and / or `premake5.lua` and re-run `doit.bat -premake`.

See https://www.sumatrapdfreader.org/docs/Our-build-system

GitHubRulesOK:
Just an interfering comment that I agree there are some odd "scroll to" behaviour differences compared to other viewers and1.7 (acrobat 9ish) should be the yardstick, however there are others more SumatraPDF specific like https://github.com/sumatrapdfreader/sumatrapdf/issues/2001 to be watched out for

Lucas-C:
OK, I'll edit the `premake5*.lua` files. By the way, it's a bit confusing to have a mix of Go, Lua & Python in the build scripts for a single C++ project...

I can move the new unit tests to `SumatraUnitTests.cpp`, but when are they executed exactly in the CI pipeline? I see this file referenced in the `test_util_files` LUA function in `premake5.files.lua`, but is it invoked by any build script ?

kjk:
There's really only Go. lua is premake's config language, so there's nothing I can do about it. The python scripts are historical artifacts. Those that are used I converted to Go. The rest is not used and left just in case.

On GitHub CI see `flgCIBuild`. It calls `buildPreRelease` which calls `build` which builds and executes `test_util` and fails if `test_util` fails.

Lucas-C:
OK, I see, thanks for the explanations!

I added a new commit

premake5.lua
```diff
@@ -652,11 +652,10 @@ workspace "SumatraPDF"
     language "C++"
     cppdialect "C++latest"
     regconf()
-    disablewarnings { "4838" }
```
kjk, line 655: There should be no changes in this file. Maybe you need to rebase

kjk (changes requested):

premake5.lua
Lucas-C: Why that ?
I introduced a dependency on MuPDF & `EngineFzUtil.h` in `SumatraUnitTests.cpp` that required those changes.

kjk: Didn't realize it had to pull in mupdf.

But looking closer, `resolve_link` doesn't need mupdf so it shouldn't use any code from it to avoid pulling it in as a dependency.

To convert string to int / float, you can use `str::Parse()` `atoi/atof` from C library https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/atof-atof-l-wtof-wtof-l?view=msvc-160

kjk: And it would probably have to be moved to a file other than `EngineFzUtil.cpp` There's no great place so can just as well go into `EngineBase.cpp`.

kjk: BTW: no tests for this functions is fine with me too. It's not that complicated.

Lucas-C: OK, I'll move that function to `EngineBase.cpp`

Lucas-C:
I think there is still an issue with `newFzLink` (_cf._ https://github.com/sumatrapdfreader/sumatrapdf/issues/2035). It should not receive a `pageNo` parameter, as `fz_link* link` already contains the destination page number. When it is called by `FzGetElementAtPos` for example, the `pageNo` that is currently provided to this function is the page number of the page **containing** the link, not the **destination** page number.

Do you mind if I fix this bug as part of this PR?

kjk:
Yes, I think you're right that `newFzLink` shouldn't need `pageNo`. Feel free to fix it.
I think I fixed the other side of this bug (i.e. https://github.com/sumatrapdfreader/sumatrapdf/issues/2035).

Lucas-C:
Great.
I added one more commit & rebased the branch to avoid conflicts.
I think it's ready to be reviewed & merged so this is no more a draft.

src/SumatraUnitTests.cpp
```diff
@@ -1,20 +1,36 @@
 /* Copyright 2021 the SumatraPDF project authors (see AUTHORS file).
    License: GPLv3 */
 
+extern "C" {
```
Lucas-C, line 4: I forgot to remove those useless imports, this is why the build broke.
I'm on it.

Lucas-C:
I fixed the missing "double to float" type casts that were failing the build.

kjk:
Thanks!

Lucas-C:
Thanks for reviewing & accepting this PR!

I did not add as part of it a mention of this change in `website/docs/Version-history.html`,
you'll take care of that?

kjk:
That page is updated when the release happens. I've noted that in `docs/releasenotes.txt`so that I don't forget.

PR #2078, opened 2021-07-29
Add faster scrolling if the cursor is over the scroll bar

magnanis:
If the cursor is over the scroll bar, scrolling with the mouse wheel will be faster.
Currently implemented with the same speed as if you were holding the alt button.
It's a feature I use a lot in other programs so I thought it could be useful.


kjk:
thanks! that's pretty clever, haven't seen that before

PR #2129, opened 2021-08-17
Minor cleanups

a1346054:
Includes spelling fixes, whitespace fixes, and similar.

Additionally, a problem was revealed regarding files pulled from upstream - they diverge a lot in terms of their executable file modes. This should be rectified by pulling those files again from upstream (or unpacking tarballs, or whichever other method was used to obtain those sources). Realistically however, those files do not belong in sumatra repository, and they make the repo needlessly large.


kjk:
I would merge the few spelling fixes that are in sumatra code.

But this touches too many files. I don't care about whitespace or file modes.

And trust me, I know better than you whether a file belongs in this repository.

PR #2151, opened 2021-08-31
Implement "fit content" zoom for images

ftk:
It works by scanning edges of a bitmap of an image and removing regions of similar colors from "page content box". Very useful for cropping page margins while reading manga or comic books.
Also included a patch which makes flicks on touchscreen work again while zoomed in.

### DIFF
--- a/src/EngineImages.cpp
+++ b/src/EngineImages.cpp
@@ -415,29 +415,20 @@ RectF EngineImages::PageContentBox(int pageNo, RenderTarget target) {
         CrashIf(x < 0 || x >= (int)bmpData.Width || y < 0 || y >= (int)bmpData.Height);
         auto data = static_cast<const uint8_t*>(bmpData.Scan0);
         unsigned idx = bytesPerPixel * x + bmpData.Stride * y;
-        return (data[idx+2] << 16) | (data[idx+1] << 8) | data[idx];
+        uint32_t rgb = (data[idx+2] << 16) | (data[idx+1] << 8) | data[idx];
+        // ignore the lowest 3 bits (7=0b111) of each color component
+        return rgb & (~0x070707U);
     };
 
-    uint32_t marginColor = 0xffffffff;
-    auto checkPixel = [&getPixel, &marginColor](int x, int y) -> bool {
-        auto rgb = getPixel(x, y);
-        // ignore the lowest 3 bits (7=0b111) of each color component and then compare it to margin color
-        const uint32_t mask = ~0x070707U;
-        if(marginColor != 0xffffffff)
-            return (rgb & mask) == marginColor;
-        // use the first pixel color as margin color
-        marginColor = rgb & mask;
-        return true;
-    };
-    auto marginReset = [&marginColor]() { marginColor = 0xffffffff; };
-
+    uint32_t marginColor;
     // crop the page, but no more than 25% from each side
-    marginReset();
+
     // left margin
+    marginColor = getPixel(0, h/2);
     for (; r.x < w / 4 && r.dx > w / 2; r.x += deltaX, r.dx -= deltaX) {
         bool ok = true;
         for (int y = 0; y <= h - deltaY; y += deltaY) {
-            ok = checkPixel(r.x + deltaX, y);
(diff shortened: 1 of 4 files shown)


kjk:
Looks decent but needs a few fixes:
* please fix the compilation error https://github.com/sumatrapdfreader/sumatrapdf/pull/2151/checks?check_run_id=3478722211
* please remove the flicking change. With the change, the page flicks even if is not scrolled all the way to the right. I would accept a change where a flick changes to next page if scrolled all the way to the right and changes to prev page if at 0 x
* cache content box on ImagePageInfo so that doesn't have to calc it every time
* every GetPage() must be matched with DropPage(page, false)

src/EngineImages.cpp
```diff
@@ -366,6 +368,125 @@ void EngineImages::DropPage(ImagePage* page, bool forceRemove) {
...
+        auto rgb = getPixel(x, y);
+        // ignore the lowest 3 bits (7=0b111) of each color component and then compare it to margin color
+        const uint32_t mask = ~0x070707U;
+        if(marginColor != 0xffffffff)
+            return (rgb & mask) == marginColor;
+        // use the first pixel color as margin color
+        marginColor = rgb & mask;
+        return true;
+    };
+    auto marginReset = [&marginColor]() { marginColor = 0xffffffff; };
```
kjk, line 431: `marginReset` is an overkill. Just write `marginColor = 0xffffffff;`
but also see a comment about speed up

src/EngineImages.cpp
```diff
@@ -366,6 +368,125 @@ void EngineImages::DropPage(ImagePage* page, bool forceRemove) {
...
+        unsigned idx = bytesPerPixel * x + bmpData.Stride * y;
+        return (data[idx+2] << 16) | (data[idx+1] << 8) | data[idx];
+    };
+
+    uint32_t marginColor = 0xffffffff;
+    auto checkPixel = [&getPixel, &marginColor](int x, int y) -> bool {
+        auto rgb = getPixel(x, y);
+        // ignore the lowest 3 bits (7=0b111) of each color component and then compare it to margin color
+        const uint32_t mask = ~0x070707U;
+        if(marginColor != 0xffffffff)
```
kjk, line 425: could speed up `checkPixel` by setting `marginColor` to `getPixel(x,y)`, which would remove a branch. Also, masking could be moved to `getPixel()`, simplifying `checkPixel` even further.

src/EngineImages.cpp
```diff
@@ -366,6 +368,125 @@ void EngineImages::DropPage(ImagePage* page, bool forceRemove) {
...
+        // use the first pixel color as margin color
+        marginColor = rgb & mask;
+        return true;
+    };
+    auto marginReset = [&marginColor]() { marginColor = 0xffffffff; };
+
+    // crop the page, but no more than 25% from each side
+    marginReset();
+    // left margin
+    for (; r.x < w / 4 && r.dx > w / 2; r.x += deltaX, r.dx -= deltaX) {
```
kjk, line 436: maybe those loops could be encapsulated into a function that takes startX, endX, startY, endY, deltaX, deltaY

ftk:
```cpp
    auto checkMarginV = [&checkPixel, &marginColor](int xstart, int xend, int ystart, int yend, int xdir, int ydir) -> int {
        marginColor = 0xffffffff; // reset margin color
        for(int x = xstart; xdir * x < xdir * (xend - xdir); x += xdir) {
            for(int y = ystart; y <= yend - ydir; y += ydir) {
                if(!checkPixel(x + xdir, y))
                    return std::abs(x - xstart);
            }
        }
        return std::abs(xend - xstart);
    };

    int left = checkMarginV(r.x, w / 4, 0, h, deltaX, deltaY);
    r.x += left;
    r.dx -= left;

    int right = checkMarginV(r.x + r.dx - 1, w / 2, 0, h, -deltaX, deltaY);
    r.dx -= right;
// then the same with x,y loops swapped
```
This looks much worse IMO

kjk:
thanks!

Issue #2054, opened 2021-07-24
Scroll not smooth under a certain pixel size


alkees75:
Hello,

I've had this problem for quite some time. I mainly read mangas in .cbz/.zip format, with images in .jpg/.png. Except that, when the pages/images are below a certain pixel size: the scroll is not fluid, either with the hand-tool or the scroll wheel. (This is more noticeable with the hand-tool). 
Example in video (First image: 822x1200 - Second image: 847x1238) 

https://user-images.githubusercontent.com/87903330/126870395-539b83ff-0572-425a-89f4-084a7315c23b.mp4

I tested several pixel sizes, and apparently all images below ~1 048 586 pixels in total have this issue.

The problem can be solved by zooming in Sumatra, or upscaling the image like I did in the video, but the quality decreases when you change the original pixel size of a page.
So I wanted to know if there was a way to fix this, while keeping the original pixel size of the image.

Here's a zip with the page used in the video, with the original size.
[Page.zip](https://github.com/sumatrapdfreader/sumatrapdf/files/6872532/Page.zip)

GitHubRulesOK:
Interesting the image "stutters" (in my machine) worse in IrfanView unless viewed at 100%
So looking at Information it appears resolution is set to 72 dpi but windows default for rendering PNG is 96 dpi This is a known 4:3 re-calculation issue that affects Images in a PDF where default unit is 72 points/inch
If I change dpi to 96 it seems to stutter less but that may just be the fact that its now in memory.

A pixel is a pixel ... so it should not matter if I change it to 96 but just check if it makes any difference on your rig? It probably does not
[pages.zip](https://github.com/sumatrapdfreader/sumatrapdf/files/6874419/pages.zip)

I saved identical scale pages at 96 dpi 001=IrfanView 002=ms paint (much larger due to bitness) 
For me all 3 seem to be equally smooth within my systems limits of smoothness.

Now back to your "issue" there is a natural stutter when scrolling in windows as the mouse jitters from line to line. If I wish to avoid windows "stepping" the mouse, I need to use the mouse auto-scroll (middle click and just move a small amount). As to why it affects one zoom lever compared to another that may be windows adjusting the dpi scaling so 96dpi at 100% may requires less recalculation compared to other zoom levels.

However I have to agree when zoomed at some ratios the flicker of each line refresh can be very annoying.

[...]

PR #2153, opened 2021-09-01
Lazy decompression of images in archives

ftk:
Greatly decreases startup time and memory consumption for archive files.
Before this patch: opening sumatrapdf with 3 large cbz files: ~80 seconds
After: ~5 seconds (in single page mode. In continuous mode there's probably no difference).
However it also increases the time to render next page. That's why I removed dimension check from DisplayModel::ShouldCacheRendering, which is probably causing other issues like #2054


kjk:
I have to think about it. I agree with the goal of speeding cbx up.

It makes opening and use in non-continuous mode faster.

However, switching to continuous mode pays the price for that and in a blocking fashion, which is not good experience. Especially on slow drives (I'm testing on a network drive, where it can take seconds to load and decompress the whole file).

It could be improved for continuous case.

For example, on cbx open we could kick off a thread to load & decompress images to get their dimensions and cache them on Engine. We don't have to decompress the whole file, the first 1 kb would be enough.

That way, by the time we need the information, it might already be there.

Still wouldn't help the case when we open in non-continuous mode, as we would need this info right away.

To help with that we could cache mediabox and contentbox in per-file settings, which would speed up opening a file the second time. That could speed up opening of other formats as well.

kjk:
Thanks! Let's give it a try.

PR #2185, opened 2021-10-11
Truncate long file names in File menu (fixes #783)

### LINKED ISSUE #783: Shorten long file names in File/recent files menu.
See https://muut.com/sumatrapdf#!/sumatrapdf/general:annoyance-with-recent-files

If file name is longer than say 60 characters, shorten it by e.g. replacing middle part with '...'

MatheusAvellar:
Fixes a long standing (albeit low priority) issue where long file names expand the File menu too much:

File menu is huge

Here's how it looks after these changes:

File menu is adequately sized

---

This isn't a trivial operation; we can't just count characters and split the string when we reach a threshold. Different characters have different widths, some glyphs are composed of multiple characters. So it's best to offload this work to the system itself.

Therefore, this was implemented via `DrawText()`, similar to how the start page truncates file names. However, the menu doesn't have a `hWnd` *or* a `hdc` readily available – and after some hours of searching, I couldn't figure out how to get a reference to either. So I settled for the next best thing: creating a new device context, and replicating the default menu font there. It's somewhat lengthy code, but it seems to do the job. Since it's my first time working this closely with Win32 API, I might have missed some easier ways to do things; I apologize in advance.

The only potential issue I can think of is that I never call `DeleteDC` for that created device context. It's a dummy device context just to measure text in a given font, so it doesn't really need to be recreated every time the menu is opened; so I believe it should only need to be deleted when the application is closed entirely, by which point I would assume that memory would be freed anyway. Please let me know if that is not the case.

Sorry for the long text! Hopefully it's been useful :)

### DIFF
--- a/src/Menu.cpp
+++ b/src/Menu.cpp
@@ -1000,37 +1000,6 @@ static bool __cmdIdInList(UINT_PTR cmdId, UINT_PTR* idsList, int n) {
 
 #define cmdIdInList(name) __cmdIdInList(md.idOrSubmenu, name, dimof(name))
 
-static HDC menuDC = nullptr;
-
-static HDC GetOrCreateMenuDC() {
-    // If we have previously created a Device Context, return it
-    if (menuDC)
-        return menuDC;
-
-    // Get the default font in use by menus (e.g. Segoe UI for latin text)
-    NONCLIENTMETRICS metrics;
-    UINT structSize = sizeof(NONCLIENTMETRICS);
-    metrics.cbSize = structSize;
-    SystemParametersInfo(SPI_GETNONCLIENTMETRICS, structSize, &metrics, 0);
-    LOGFONTW menuFont = metrics.lfMenuFont;
-
-    // Create device context
-    HDC hdc = CreateDC(TEXT("DISPLAY"), NULL, NULL, NULL);
-    // Create HFONT for use in the device context
-    // TODO: is there a more succint way to convert LOGFONTW -> HFONT?
-    HFONT hFont = CreateFont(menuFont.lfHeight, menuFont.lfWidth, menuFont.lfEscapement, menuFont.lfOrientation,
-                             menuFont.lfWeight, menuFont.lfItalic, menuFont.lfUnderline, menuFont.lfStrikeOut,
-                             menuFont.lfCharSet, menuFont.lfOutPrecision, menuFont.lfClipPrecision, menuFont.lfQuality,
-                             menuFont.lfPitchAndFamily, menuFont.lfFaceName);
-    SelectObject(hdc, hFont);
-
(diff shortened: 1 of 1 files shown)


Qynklee:
I think it's very wel in some cases but in other cares, filenames with number like ABC_1, ABC_2... so hard to choose.
Truncate long name is good but truncate long name with small pop-up when mouse hover to show complete filename is better.
😄 
Ps: I'm a new in this project, trying and hope that can improve it 😄

MatheusAvellar:
@kjk

> I think limiting the string by character count is better.

It's definitely a possibility; I'd seen [this SO answer](https://stackoverflow.com/a/6043797/4824627) stating that UTF-8 was built with this in mind, and makes it easy to check if you're splitting a character in half or not. There were a couple reasons I went for the other approach, though:

1. Wide characters can make the list look very jagged. A file name with thin characters can ends up significantly smaller than one with wide characters. For example, 'i's and 'W's in the screenshot below. The bounding rectangle does not have this issue.
2. Some characters are made up of multiple codepoints (e.g. emojis come to mind, but possibly also chars in other scrips). When splitting by a fixed threshold, they can end up broken. For example, file (7) below should end with a '👉', but ends with a tofu instead. Using the system's own `DrawText()` to truncate text fixes this.

These are set to truncate at 60 chars:

Recent files list with file names truncated at 60 characters

---

> Have you checked how other programs on Windows (e.g. Notepad++ or Adobe Reader) handle this?

That's a great idea! It didn't occur to me.

Notepad++ and Sublime Text 3 don't seem to truncate filenames at all 🥴

Visual Studio does what you suggested: they add an ellipsis in the middle of the file name. It also seems like it's generating a tofu for file (2), however their list is not as jagged; there has to be some sort of bounding box at play here:



Didn't check Adobe Reader because I'm in a bit of a rush at the moment, but I could try looking for more examples in the wild tomorrow to get a feel of the industry standard on this.

Either way, I support the simpler idea, or perhaps a best-of-both-worlds approach. Eager to hear your input on it :)

kjk:
Perfect can be enemy of good.

The simple implementation is to cut WCHAR characters from the middle.

It's true that some glyphs are represented as 2 WCHAR characters and cutting just one of them would generate imperfect text but that would be very rare. I'm not even sure if such characters are allowed in file names.

So a slightly more complicated implementation can ensure to preserve those 2 WCHAR glyphs.

And even more complicated implementation would measure text to decide how many characters to cut instead of just setting max characters.

I'm fine with the simple implementation.

MatheusAvellar:
Implemented the simpler method, here's how it truned out (truncating at 70 chars):

Recent files list with file names truncated, from the middle, at 70 characters

PR #2189, opened 2021-10-13
Fix #1553 - color bug in "Page:" and "Find:" toolbar labels

### LINKED ISSUE #1553: The toolbar "Page:" and "Find:" color will be same TextColor in setting file.
FixedPageUI [
	TextColor = #ffffff
	BackgroundColor = #263238
	SelectionColor = #3399ff
	WindowMargin = 2 4 2 4
	PageSpacing = 4 4
]

It will display "Page:" and "Find:" as white under this config, and it's undistinguishable from toolbar background color.

It's reproduces in recent new svg toolbar daily builds.

echedey-ls:
Also, fixed a typo.


GitHubRulesOK:
@echedey-luis-alvarez 
I am not a developer either :-) and understand the code less, but do understand that a minority of users moan about many random issues that the majority accept, Thus you cant please, all of the zero paying clients, all of the time. 🙂

PR #2230, opened 2021-11-02
Updated translation to pt-br

Felipefpl:
Aside the updated translation i wanted to report that the latest installers of sumatra are installing the program in the C:\Users\xxxxxxx directory instead of the program files folder, i'm on win7 x64 and installed the 32 bit version.


kjk:
Thanks, but translations are maintained in https://www.apptranslator.org/
This file is automatically generated from there.
To update translations please use https://www.apptranslator.org/app/SumatraPDF

Installation in `Users` directory is an intentional change.

kjk:
See https://www.sumatrapdfreader.org/docs/How-to-contribute-translation

PR #2260, opened 2021-12-02
add relative path, and If you change your computer, the path is repai…

DesolationIsland:
Repair the file path by determining the machine name and relative path information.


kjk:
I don't understand the use case for this change.

If something doesn't happen often, we don't want to complicate the code to support it.

Issue #606, opened 2016-09-01
Please add smooth touchpad support


vvRFvv:
I have a Thinkpad with a synaptics touchpad and the newest version of windows. It would be nice to have smooth touchpad scrolling and zooming (not in steps) as it is already possible when using the touchscreen and sumatra.

kjk:
I don't know what "smooth scrolling and zooming" means in this context.

Zooming will always happen in steps because it's very expensive operation.

Scrolling already happens in small increments (assuming contiguous layout), which should be true for mouse, touchpad and any other device that generates scroll events.

vvRFvv:
Please take a look at the video. First I use the touchscreen to scroll and zoom, then the trackpad. Especially the zooming with the trackpad has huge steps.

[sumatraScrolling.zip](https://github.com/sumatrapdfreader/sumatrapdf/files/451715/sumatraScrolling.zip)

myphs:
Yes, this already works, but Sumatra is ignoring the input.
You also already can scroll smoothly when you hold the right mouse button and move the mouse, but then you are forced to release the RMB and move to another place to scroll again. On a laptop this is very unconvenient.
When using precision touchpad the two finger scroll should be smooth as well then.
I would very appreciate if this feature could be implemented.

GitHubRulesOK:
SumatraPDF is not ignoring any scroll input it is respecting the inputs from Windows telling it to move # lines of text

Touchpad or mouse scrolling (much the same as font sizes) is tied to the OS (Windows settings) there is no 0.01 line granularity it is simply # of lines in a step range of 1 to 99 (or page by page)
Please write to Bill G and his team to change that.

![image](https://user-images.githubusercontent.com/12754913/74693231-c81ca280-51e2-11ea-9db3-7d01b7d6245d.png)


SumatraPDF uses an engine that is similar in reaction to similar document apps (check out your word Pad etc. and yes large programming teams can bypass that limitation in most paid for applications)

The next best option is Auto scroll drivers which are common on the middle button of a mouse but not always found on touchpads Please write to Synaptics or others to add that option if it is missing.

The third option is coasting not usually found on mice but may be offered on some touchpads however it is normally tied to the default number of full lines as above.

Finally, we have the fall back of smooth grab'n'drag as described by Myphs Which, if you have enabled vertical scrolling on your touchpad, can be achieved within the hardware limits by holding down the right button and sliding a finger up and down the slider zone.

[...]

PR #2343, opened 2022-02-13
Implement support for smooth scrolling.

msc94:
This commit adds support for smooth scrolling. ~~There is a configurable option in the settings dialog to enable smooth scrolling.
If this option is checked, when a scroll event occurs, a timer is started that gradually scrolls to the desired y offset.~~

~~I still need to figure out how to persist the smooth scrolling setting into the config file.
Some pointers in the right direction would be cool!~~

This PR would close https://github.com/sumatrapdfreader/sumatrapdf/issues/606

### DIFF
--- a/src/Canvas.cpp
+++ b/src/Canvas.cpp
@@ -57,12 +57,12 @@
 #define MW_SMOOTHSCROLL_TIMER_ID 6
 
 // Smooth scrolling enabled?
-static bool gSmoothScrollingEnabled = true;
+static const bool gSmoothScrollingEnabled = true;
 
 // Smooth scrolling factor. This is a value between 0 and 1.
 // Each step, we scroll the needed delta times this factor.
 // Therefore, a higher factor makes smooth scrolling faster.
-static double gSmoothScrollingFactor = 0.2;
+static const double gSmoothScrollingFactor = 0.2;
 
 // these can be global, as the mouse wheel can't affect more than one window at once
 static int gDeltaPerLine = 0;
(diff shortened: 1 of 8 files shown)


msc94:
Zooming while scrolling is also still buggy. I need to fix that.

kjk:
1. Accidentally checked in `src/SumatraPDF.aps`. I shouldn't have  checked it in at all, I removed it on main branch. You might need to rebase.
2. I'm not sure I want this to be controlled by a setting. If smooth scrolling is better than regular, then it should just be enabled. For now, could you make it controllable via a global variable (for easy testing) and revert the changes related to settings.
3. Even if there will be a setting, it'll only be from settings file. Number of settings in UI is on purpose very limited
4. For future reference: to add settings, you need to follow process described at
https://www.notion.so/How-we-store-settings-478e6d9511ae4d24b4792da4e8809c1f

msc94:
I added a global feature flag in Canvas.cpp and removed all changes related to settings and the UI now.

a-silent-message:
I've been needing this feature for forever, yet to test this out, but thank you @msc94

geekrelief:
I tested this out.  Smooth scrolling definitely improves the reading experience. Mouse wheel scrolling is working well, but when I use the middle mouse button to scroll there's some jank/stutter when scrolling quickly. It'll scroll smoothly then jump a little bit. Setting `Selection.h: SMOOTHSCROLL_DELAY_IN_MS`  to `USER_TIMER_MINIMUM` helps but doesn't eliminate it.

niv-dev:
Hi, is there anyway that this can be implemented with horizontal scroll aswell?

PR #2580, opened 2022-05-17
select first annotation when delete

kimim:
when delete an annotation each time, the delete button will disappear, you need to re-select an annotation to delete.
this PR will select the first annotation of the remaining, you can continue to press delete button till all annotations are deleted.


kjk:
thanks

Issue #2565, opened 2022-05-11
Load files in a background thread


kjk:
Currently if a file is on a slow network drive or if it's a complicated file, it'll block main UI thread while it's loaded and parsed.

Load the file in a background thread, show a `loading in progress` message.

Will also help when restoring a large number of files.

kjk:
Implemented with `LazyLoad = true` advanced setting.

Issue #1497, opened 2020-03-15
Lazy load tabs to speed up the startup time


fatjing:
The startup performance lacks severely if the program is configured to load the previous session on start, especially when a lot of tabs need to be restored.
It'd be nice to have only the foreground tab restored at startup, and lazy load those background tabs upon switching.

jddcef:
It loads pretty fast for me the last year or so, using the 3.5 etc pre-release versions.

kjk:
I've implemented loading on background threads, so slow documents no longer block the main thread (i.e. you get fast startup time).

jddcef:
Just realised, why is it called a "pre-release", if it's actually AFTER the last release. 

A pre-release sounds older, as if it's before (pre) the actual release, therefore outdated!

p.s. the website's page is missing links to download on https://www.sumatrapdfreader.org/prerelease

kjk:
Well, it's before the next release. But I see how this might be confusing.

> p.s. the website's page is missing links to download

You must have JavaScript enabled.

hpvd:
follow up on optimizing startup time incl. approach see https://github.com/sumatrapdfreader/sumatrapdf/issues/4803

PR #2590, opened 2022-05-21
Support lazyload when restore session

ultranity:
Lazy load session tabs when startup untill switch to the target tab for lower mem usage and speed up.

When select other tabs, there will be a highlight message "Please wait - rendering..." before the document is loaded (using ReloadDocument).

User may change lazyloadWhenRestore setting to disable it.

In the future we may add scheduled load in background (but will increase mem usage so better make it optional) #2565, or add a DisplayModel with lazyload support for better code structure.

#1497

### DIFF
--- a/src/Settings.h
+++ b/src/Settings.h
@@ -329,6 +329,8 @@ struct GlobalPrefs {
     // if true and SessionData isn't empty, that session will be restored
     // at startup
     bool restoreSession;
+    // if true, lazyload session tabs on startup
+    bool lazyloadWhenRestore;
     // ISO code of the current UI language
     char* uiLanguage;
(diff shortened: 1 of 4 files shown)


kjk:
Looks interesting.

Could you re-submit against `next` branch?

I'm very close to releasing 3.4 and master is currently the 3.4 branch where I don't want to make changes unless absolutely necessary to fix a serious bug.

I've been doing a lot of refactoring on `next` branch so it diverged from master quite a bit.

For now just make it configurable via a global variable, not settings, like:
```
static bool gEnableLazyLoad = true;
```
It might just be the only behavior. Either way, it's easy to add a setting later.

PR #2591, opened 2022-05-23
support lazyload (next )

ultranity:
>Lazy load session tabs when startup untill switch to the target tab for lower mem usage and speed up.
>
>When select other tabs, there will be a highlight message "Please wait - rendering..." before the document is loaded (using ReloadDocument).

#2590
Rebase on next branch.
~User may change lazyloadWhenRestore setting to disable it.~
Add a global variable static bool gEnableLazyLoad = true in settings.h

>In the future we may add scheduled load in background (but will increase mem usage so better make it optional) #2565, or add a DisplayModel with lazyload support for better code structure.
>
>#1497

Still, it will be nice to move all loading to be async, for more fluent loading UX, but I wish there could be another setting to select whether we need load all tabs, or only load when necessary (like this ver.). 

And one step further we may oneday support to freeze some tabs to constrain mem usage like Chromes, a not so necessary feature lol.

### DIFF
--- a/src/Settings.h
+++ b/src/Settings.h
@@ -287,6 +287,9 @@ struct SessionData {
     int sidebarDx;
 };
 
+// if true, Enable lazyload session tabs on startup
+static bool gEnableLazyLoad = true;
+
 // Most values on this structure can be updated through the UI and are
 // persisted in SumatraPDF-settings.txt
 struct GlobalPrefs {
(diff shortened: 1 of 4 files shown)


kjk:
Thanks.

2 notes:
* it doesn't work for drag&drop (I think because we switch to each new tab)
* we should clear background when switching to a not-yet-loaded tab. When a document takes a long time to load (can happen with e.g. complex mobi / epub files), we show the stale image, which is confusing

I agree about async loading.

Shouldn't be that hard: I think it can be done by:
* split `LoadDocument` into two parts:
  * `LoadDocument` and e.g. `LoadDocumentRest`
* in `LoadDocument`
  * show a notification `Loading ${filePath}`
  * on a thread do `CreateControllerForEngine()\CreateControllerForFile()` (move this work from the middle of `LoadDocument` to be the first thing
  * when loading fails, show `Loading failed` notification
  * if loading ok, use `uitask::Post()` to schedule running remaining of the code (i.e. `LoadDocumentRest`) on ui thread

PR #2758, opened 2022-06-27
Fix clobbered text in enter password dialog

harig:
Use the correct type for the filename text which is displayed.


kjk:
Thanks!

PR #2828, opened 2022-07-27
Update TRANSLATORS with Corsican

Patriccollu:
Adding [Patriccollu](https://www.apptranslator.org/user/pat_sanmar) as **Corsican** translator for numerous contributions since 2018.


kjk:
Thanks!

PR #2881, opened 2022-08-20
Fix overlapped rendering of tabs

harig:
When the window width was stable and tabs were removed/closed and
opened, the new tabs were rendered in an overlapped way. Fix this by
doing the layout of the tabs in such a case.

Issue # 2806


kjk:
Thanks!

PR #3022, opened 2022-10-13
Bookmark font

Shan-Zhou:
The current version use system default font (e.g. Segoe UI) for bookmarks, which might look weird when the language of OS doesn't match well with the language of bookmarks. Besides, depending on the size and resolution of the screen, users might also want to adjust the font weight. It would be good to support a file-level adjustment of bookmark font, but a quick fixing could be using a user-defined font that contains large enough character sets.

Support user-defined font and adjustable font weight by loading font name and weight offset (weight = system default weight + offset) from the setting file and overwriting the default font setting.

Example:

> default font:
> ```
> TreeFontSize = 0
> TreeFontWeightOffset = 0
> TreeFontName = automatic
> ```
> user defined font 1:
> ```
> TreeFontSize = 30
> TreeFontWeightOffset = 200
> TreeFontName = Source Han Serif
> ```
> user defined font 2:
> ```
> TreeFontSize = 30
> TreeFontWeightOffset = -200
> TreeFontName = Ysabeau
> ```


kjk:
Thanks!

PR #3089, opened 2022-11-12
Flat borders for toolbar textboxes

adoriandoran:
The appearance of the toolbar improved significantly with the new outline icons. However, the 3D borders of the textboxes for page number and search are not suited for a modern style. I replaced the 3D borders with flat borders, which better integrate with the icon style.

Here are a few screenshots showcasing this change:

On Windows 11

On Windows 7


kjk:
Thanks!

PR #3144, opened 2022-11-29
Correct file icons in Windows Explorer (fixes #3074)

### LINKED ISSUE #3074: Associated files have wrong icons in Windows Explorer
After updating from 3.3.3 to 3.4.6, the PDF files in Windows Explorer had icon of the SumatraPDF program, instead of the "PDF document" icon.

I manually fixed it in the Registry:
```
[HKEY_CLASSES_ROOT\SumatraPDF.pdf\DefaultIcon]
```

- replaced: `C:\Program Files\SumatraPDF\SumatraPDF.exe`
- with: `C:\Program Files\SumatraPDF\SumatraPDF.exe,1` (no quotes needed)

Though, it might be advisable to add quotes, in case the path contains a `,` character:
`"C:\Program Files\SumatraPDF\SumatraPDF.exe",1`

----

I pinpointed the introduction of this bug to these commits:
* https://github.com/sumatrapdfreader/sumatrapdf/commit/2c076e273d0ff899f414705b1fc7a9f56a57815e#diff-2b3b54a31776cd99b879c09874d3fb92b4b13b62091b72e36f04b4371e03f9a2R176
* https://github.com/sumatrapdfreader/sumatrapdf/commit/172afb4967e6cedc185716f807089a4f02c26a61
* https://github.com/sumatrapdfreader/sumatrapdf/commit/815bfbdbb6bd85a31b004e169375725a307cc1b8#diff-2b3b54a31776cd99b879c09874d3fb92b4b13b62091b72e36f04b4371e03f9a2

The mistake is that in Shell associations, the icons from the executable are 0-indexed, from top to bottom then from left to right, and including the program icon (which is 0, the default). Maybe the Windows API works differently, hence the confusion.

So with resources embedded in the 3.4.6 exe:
SumatraPDF exe icons

indexes are: **0** for program icon, **1** for PDF (and other) files, **2** for ePub files, and **3** for cbX files.

vlakoff:
Here we are simply dealing with the 0-based icon indexes, not with resource identifiers.

Also enclosing the path in quotes, in case it contains a "," character.

### DIFF
--- a/src/RegistryInstaller.cpp
+++ b/src/RegistryInstaller.cpp
@@ -165,12 +165,13 @@ bool RegisterForOpenWith(HKEY hkey) {
         // Per https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-extracticona
         // ",${n}" => n is 0-based index of the icon
         // ",-${n}" => n is icon with resource id
-        char* iconPath = str::JoinTemp(exePath, "");
+        char* iconPath;
         if (str::Eq(ext, ".epub")) {
-            iconPath = str::JoinTemp(exePath, ",-3");
-        }
-        if (str::Eq(ext, ".cbr") || str::Eq(ext, ".cbz") || str::Eq(ext, ".cbt") || str::Eq(ext, ".cb7")) {
-            iconPath = str::JoinTemp(exePath, ",-4");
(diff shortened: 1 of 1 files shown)


kjk:
Thanks!

PR #3233, opened 2023-01-19
Added command line option to disable auto rotation when PDF is not in portrait mode

### LINKED ISSUE #3073: Unable to print in landscape format?
Somehow I'm unable to print any pdfs in landscape format in SumatraPDF. It happen with any pdf, tested it with two different printers and two different SumatraPDF versions. It doesnt matter what i select in the printer settings (portrait or landscape format), SumatraPDF will always print it portrait.
Printing in landscape format works fine in Adobe Reader or Browser Add-Ins (e.g. Chrome or Edge).

version of Sumatra: tested on 3.1.2 and 3.4.6 (newest)
version of the OS: tested on Windows 10 & 11
name of the printer: tested Zebra GK420d & Canon LBP223dw
can you open and see the file in Sumatra?: yes
what exactly happens, are there any error messages?: see above, no errors

priyeshshah11:
Adds a new argument `disable-auto-rotation` to override auto rotation when the PDF is not in portrait mode & gives preference to the user specified orientation.

Fixes #3073 - We get this issue when we are trying to print a portrait 6" X 4" PDF in landscape mode.

### DIFF
--- a/src/Flags.cpp
+++ b/src/Flags.cpp
@@ -527,7 +527,7 @@ void ParseFlags(const WCHAR* cmdLine, Flags& i) {
         }
         if (arg == Arg::PrintSettings) {
             // argument is a comma separated list of page ranges and
-            // advanced options [even|odd], [noscale|shrink|fit] and [autorotation|portrait|landscape]
+            // advanced options [even|odd], [noscale|shrink|fit] and [autorotation|portrait|landscape] and disable-auto-rotation
             // e.g. -print-settings "1-3,5,10-8,odd,fit"
             i.printSettings = str::Dup(param);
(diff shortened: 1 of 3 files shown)


kjk:
Thanks!

priyeshshah11:
@kjk Thanks for merging this. However, we are still having issues printing PDFs in landscape mode, the job gets sent to the printer but then disappears straight away without printing. 

I noticed that our original PDF size is ~1MB, which then increases by 4x when `sumatraPDF` sends it to the printer & increases by 8x if we specify an orientation in print settings. Any ideas/suggestions?

kjk:
Well, the printing code in Sumatra sends rasterized pages, hence the big output.

Not sure why an orientation would increase the size further.

PR #3278, opened 2023-02-16
Fix spelling error of "directory" in translations

guemax:
I found a small typo in translations.txt while reading the latest changes of the SumatraPDF source code.

### DIFF
--- a/src/docs/translations.txt
+++ b/src/docs/translations.txt
@@ -13224,7 +13224,7 @@ br:Abrir Pasta em Directory &Opus
 cy:Agor Cyfeiriadur yn Directory &Opus
 de:Verzeichnis in Directory &Opus öffnen
 it:Apri cartella in Directory &Opus
-nl:Open Map in Direcotry &Opus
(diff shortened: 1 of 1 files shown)


kjk:
Those are actually extracted from code and translations are downloaded from https://www.apptranslator.org/app/SumatraPDF

guemax:
So I should have fixed the typo in AppTranslator rather than in the extracted file, am I right?

P.S. I have already seen that this problem has been fixed, thanks! 👍

kjk:
The English strings are in the source code, marked with `_TRN(` so they have to be changed like this: https://github.com/sumatrapdfreader/sumatrapdf/commit/1658028bd113128084d72c3f32d676c2bd8e5e0d

They are extracted from the source periodically and uploaded to apptranslator for translations.

Translations have to be changed in AppTranslator and they are periodically downloaded and latest version is stored in src/docs/translations.txt

guemax:
Thank you for your quick reply!

PR #3321, opened 2023-03-03
Increase Max Copies when Printing

bierpub:
When printing a PDF, increase the maximum number of copies from 999 to 9999. Currently anything over 1000 will revert back to 1 copy. I would like to have the option to queue up more than 1000 at a single time. I tested and this works for my purposes going up to a max of 9999 copies.

`SumatraPDF-prerel-64.exe -print-to "PrinterXYZ" -print-settings "9999x" -silent file.pdf`

### DIFF
--- a/src/Print.cpp
+++ b/src/Print.cpp
@@ -1122,7 +1122,7 @@ static void ApplyPrintSettings(Printer* printer, const char* settings, int pageC
             advanced.rotation = PrintRotationAdv::Landscape;
         } else if (str::EqI(s, "disable-auto-rotation")) {
             advanced.autoRotate = false;
-        } else if (str::Parse(s, "%dx%$", &val) && 0 < val && val < 1000) {
+        } else if (str::Parse(s, "%dx%$", &val) && 0 < val && val < 10000) {
             devMode->dmCopies = (short)val;
             devMode->dmFields |= DM_COPIES;
         } else if (str::EqI(s, "simplex")) {


GitHubRulesOK:
hmmm
 Print trays usually are 500 max so physical printers need 2 trays (possible) to reach 1000 but virtual printout of more than 500 is usually ill advised as the file size bloats to slow or massive and could easily reach the practical limit of under 2GB.pdf. not my place to say what a limit should be but 1000 seems very reasonable. Although pages over 4000 are fast when vector objects under 2MB as I have shown in the past but print as image is not advisable at such levels. What would be a practical Usage?

bierpub:
I am printing to a label printer that will print out 1 label at a time as necessary, and will continue printing hundreds or thousands during a single run. The labels are usually 4x6 inch or smaller. Currently I queue 999 copies at a time and keep resending to the printer as necessary. It would be more convenient to send 9999 copies and not have to re-send as often. Sending more labels also avoids problems during a temporary network outage where the print server might not be able to send to the printer for a short period of time. With more labels already queued up before the outage, there is less chance of running out.

I also generated some PDF's and my file sizes are much smaller. I used Sumatra to print to a PDF printer. It started as a single page pdf that was printed with 9999 copies using `-print-settings "9999x"`.
image

4x6 inch label
image

8.5x11 inch sheet
image

kjk:
Thanks but I'm going to decline this.

There's an easy work-around: execute printing multiple times.

The limit is there for a reason: to prevent mistakes. And it's probably too big anyway but not going to make it bigger.

PR #3374, opened 2023-03-21
When dragged to a different window and released, add tab to that window

avorobey:
The change from ReloadDocument() to LoadDocument() is because the former checks assumptions on win->ctrl and tab->ctrl that aren't true in our novel use case. And anyway, LoadDocument with args->reuseTab is exactly the right thing to do in our case; and is what ReloadDocument() would have done if we hackily adjusted win->ctrl=nullptr for it.

### DIFF
--- a/src/Tabs.cpp
+++ b/src/Tabs.cpp
@@ -114,11 +114,11 @@ void MigrateTab(WindowTab* tab, MainWindow* newWin) {
     MainWindow* oldWin = tab->win;
     RemoveTab(tab);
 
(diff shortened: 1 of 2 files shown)


GitHubRulesOK:
cannot say more than "impressed I am, Padawan" Old forgetful me, likes to think of you both as young Masters of the Force

kjk:
Thanks!

kjk:
@GitHubRulesOK yeah, only 9 years to close it :)

PR #3388, opened 2023-03-26
Preserve and use destination fragment in inter-file PDF links

### LINKED ISSUE #2635: Respect page argument in links to files
PDF links from one local PDF file to another PDF do not go to the right page of the target file. Sumatra successfully opens the target PDF, but it does not go to the right page. Instead it always goes to page 1. I would rather expect that it would jump to the right page in the target PDF.

### Background:
We are creating our PDFs using Latex and are intensively using external cross-references. These are hyperlinks from one PDF to another one, created with the XR-Hyper Latex package.

### Versions:
We like Sumatra very much. It is our standard PDF viewer. Until Version 3.1.2 our external PDF links worked very well. But all newer versions (up to most recent 3.4.3) exhibit the issue. 

Please also note that other PDF viewers (such as Acrobat) do not have problems.

### Reproduce:
The issue is easily reproduce with both attached PDF files. 
[source.pdf](https://github.com/sumatrapdfreader/sumatrapdf/files/8798656/source.pdf)
[target.pdf](https://github.com/sumatrapdfreader/sumatrapdf/files/8798657/target.pdf)
Open the source.pdf and click the link. It will open target.pdf. 
But it always starts at page 1, instead of the right target page 2.

avorobey:
Fixes #2635.

### DIFF
--- a/src/EngineBase.h
+++ b/src/EngineBase.h
@@ -120,21 +120,32 @@ struct PageDestinationURL : IPageDestination {
 
 struct PageDestinationFile : IPageDestination {
     char* path = nullptr;
+    char* name = nullptr;
 
     PageDestinationFile() = delete;
 
-    PageDestinationFile(const char* u) {
+    PageDestinationFile(const char* u, const char* frag) {
         CrashIf(!u);
         kind = kindDestinationLaunchFile;
         path = str::Dup(u);
+        if (frag) {
+            name = str::Dup(frag);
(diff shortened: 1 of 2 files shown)


kjk:
Thanks!

For future reference: no need for null check around str::Dup(), str::Free(). str::Dup() returns null on null and str::Free() is just free(), which is no-op on null.

Issue #2900, opened 2022-08-26
Mouse 3 (middle mouse) doesn't close tab?


nabijaczleweli:
Clicking on the tab with button 3 doesn't close the tab; this is counter to my expectations, given that in every program I can think of that's exactly what happens.

Chasing the little X in the corner is rather annoying :)

nabijaczleweli:
Oh, actually, this works *except* if you click on the text. But the text is most of the tab, so.

shuheque:
@nabijaczleweli Also find this annoying, I remember I can close tabs w/ middle button in old versions that has NO tooltips on the tab file names, maybe I'm wrong?

GitHubRulesOK:
Hardware/user action will have a bearing if you trigger the tooltip/context/error message on click down then the focus is not on the tab hotspot to close on the mouse up event, as its absorbed in any other action. usually move away a repeat slowly without dragging tabs etc will help

nabijaczleweli:
idk, haven't used old versions

GitHubRulesOK:
@nabijaczleweli 
[try pre-release 3.5.15380 + ](https://www.sumatrapdfreader.org/prerelease)

nabijaczleweli:
yep, works for me

Issue #880, opened 2017-09-04
Middle click does not close tabs reliably


noperator-zz:
In order to close a tab with the middle click button, 'TabBarProc' (Tabs.cpp) must receive a 'WM_MBUTTONDOWN' message followed by a 'WM_MBUTTONUP'. The 'WM_MBUTTONDOWN' handler calls 'TabNotification(..., T_CLOSING, ..., ...)' which puts the clicked tab into tab->xClicked. Then the 'WM_MBUTTONUP' handler closes the tab.

A problem occurs when multiples tabs are middle clicked in quick succession. 'TabBarProc' misses the 'WM_MBUTTONDOWN' event so tab->xClicked is never set, which causes 'WM_MBUTTONUP' to do nothing.

I don't know how 'TabBarProc' is called, but the best solution would be to make sure it never misses messages, as I'm sure that causes other unwanted effects.
Alternatively, the functionality of 'WM_MBUTTONDOWN' can probably just be copied into 'WM_MBUTTONUP'

imsoenthused:
Middle-click also has an issue caused by any high enough DPI sensor mouse. The click to close tabs absolutely will not register unless the mouse is in the same EXACT position for the entire process. With high DPI mice, the act of clicking will cause an almost imperceptible movement of the mouse but the high-resolution sensor registers it causing the attempt to close to completely fail.

GitHubRulesOK:
@kjk 
I observe in my browser I can click middle down at one point within a tab move mouse elsewhere in the zone then release while close the tab, but cant do that in Sumatra.
I know you worked hard recently on tooltips but can that constant detection not also be used to boundary the up middle mouse ?

GitHubRulesOK:
@kjk 
just a reminder that current middle click seems not comparable to previous see https://forum.sumatrapdfreader.org/t/closing-tab-with-middle-click/3421/7

PR #3391, opened 2023-03-27
Improve middle button closing of tabs

avorobey:
Middle button closing of tabs is flaky b/c even a slight movement between press and release causes tabBeingClosed=-1 in WM_MOUSEMOVE. That's meant for canceling a left-button close when we move away from X before releasing, but it also affects middle-button badly. Remove that and instead check more stringently when left button releases. This restores middle button to its glory and as a bonus when we left-press, move away then come back to the X then release, the tab closes after all - the normal Windows behavior.


GitHubRulesOK:
related https://github.com/sumatrapdfreader/sumatrapdf/issues/2900 https://github.com/sumatrapdfreader/sumatrapdf/issues/880

avorobey:
@GitHubRulesOK ah, thanks for linking, I think I saw this originally in #2900 but then forgot where it was.

kjk:
Thanks!

PR #3394, opened 2023-03-28
Theme support: tabs, theme-switching key

avorobey:
The existing Theme structs have more than just colors for tabs (pen width, whether or not to draw the X circle, etc.). Also the stuff in AppColors.cpp doesn't do anything nontrivial with tab colors (unlike main doc colors which several global settings can influence). Taken together, it seems a good idea to let go the AppColor intermediary for tab colors and get them from Theme directly. This commit does that, plus add a Ctrl-Shift-D key to cycle the theme. This is just for convenient testing, I don't know what the key should be, and there probably should be a separate mechanism in the menu.

It is NOT a complete theming support, the main colors are trickier and will come separately. I'm not yet sure whether to leave the AppColor intermediary for them or incorporate the logic in Theme.cpp.

The stuff in Theme.cpp was a bit bitrot, but just a bit. Fixed a compile error by changing char* name to const char*, and changed some of the actual colors in the light theme to newer values in AppColors.cpp.


kjk:
Thanks!

Looks like Accelerators.cpp got re-formatted.

Tip: run `.\doit.bat -format` to format all the source code with `clang-format` (can be installed as part of Visual Studio or by installing `llvm`). Good idea to run before checkin but nbd: I run it ocassionally.

`AppColors.[h|cpp]` predates theme so it make sense to eventually get rid of and just use themes.

`AppColors` has some logic to use system values but in retrospect it was a mistake. People will report visual bugs due to custom Windows theme and I can't reproduce them. It's better to have fixed colors so that everyone gets the same look.

We should completely remove (deprecated) useSysColors option.

There's also a question how it should interact with invert colors (`i`). I'm guessing most people use Invert Colors as the best approximation of dark theme. Then again, the inverting of document colors is different than changing colors of UI elements, so maybe that should be separate.

The tricky part is ensuring all windows constrols (list boxes, tree views, toolbar, menu) respect the theme colors.

In some cases it requires switching to custom drawing because the controls don't provide enough customizability out of the box. Menu for sure, possibly tree view to get everything right.

kjk:
There's also caption (title bar): Windows only provided API to change it in win 11.
There's a way to take over painting that part by extending client area into caption, which is how we do tabs in `Caption.cpp`, but currently we don't do it in non-tab mode and in tab mode we don't cover all the area i.e. close / minimize / maximize and some of the upper and right parts are painted by Windows.

We can either punt it as known limitations or improve `Caption.cpp` to cover 100% of caption and do the same in non-tab mode.

avorobey:
@kjk thanks, very useful guidance.

The .\doit.bat -format thing - will it also convert CRLF back to LF for me? git is bugging me about this constantly. I'm writing code under Windows for the first time in 15 years or so, and let me tell you, I did not miss the CRLF thing. 

I hear you about the windows controls and the title bar, I did not realize those were the trouble at all. We'll make it work one way or the other.

I'm glad useSysColors  can go. What about gGlobalPrefs->fixedPageUI.{backgroundColor,textColor}? That's the third nontrivial part of AppColor (besides useSysColors and invertColors). For black-and-white PDFs those are basically the two main colors on the screen, right? Going to a dark theme ought to invert them, but if the user made custom values for them in the settings or command line arguments, how will that work?

kjk:
re: git, try setting the git config:
```
git config --global core.whitespace cr-at-eol
git config --global core.autocrlf false
```
I think only `core.autocrlf` is relevant but showing my git config.
I don't think `clang-format` changes newlines. In my VS I see lines end with `CRLF`.

re: fixedPageUI colors and all other current color settings
Long term they all belong in theme i.e. theme colors should have the equivalent of those settings.

The question is: how to handle the transition?

One option: when themes have replacement values, after loading the settings over-write default theme with those values and set them to some "impossible" value like -1, to prevent over-writing in the future. Then after few releases remove those settings.

Which brings bigger question: how to allow customization of themes, adding new themes?

First cut could be just to ship 2 themes: current (call it light) and a new dark theme, no customization.

In which case we would keep current customization option and they would over-write few colors in the theme.

One option for customization is to use the current setting systems and allow people to add new named theme. Our 2 themes would be baked in the executable, say names "standard-light" and "standard-dark" so that they don't have to pollute settings file.

We would publish their definition in the docs so that people could copy & paste definition to settings file under a new name and modify the colors they want (so that they don't have to start from scratch).

That would also be fairly easy way for people to share the themes.

I'm also fine with just ditching those current customization option and telling people: sorry for loosing the setting, here's the new way to achieve the same thing (in the docs, release notes etc.)

GitHubRulesOK:
@avorobey 
> let me tell you, I did not miss the CRLF thing.  

🤣  I have to deal with all three \r ,\n, \r\n

@kjk I think for tabs let the use use a rgb triplet like  "Gradient Colours" if they bork them, set a default fail back

PR #3402, opened 2023-03-29
utf-8 for global users and mild modifications

ingyer-ks:
Dear,
I tried to build your pure code but it didn't work.
I applied /utf-8 option to cl.exe, AutoFreeStr being char* by calling Get() and disabled debug  so unnecessary debug libraries are excluded.
There are still problems on signfile, but anyway sumatrapdf.exe has been built.
Please check my commits.
Thank you for making this.


kjk:
1. To modify vs solution you need to modify `premake5.lua` file and re-generate solution. See https://www.sumatrapdfreader.org/docs/Build-system

2. You don't need to build signfile. It's not used anywhere

3. I'm concerned that just setting utf-8 might have unwanted consequences

4. There are unrelated things packed into this change so I have to either apply everything or nothing

Let me think about /utf-8. If I decide to do it, I'll make the change.

Can you list the files that fail to build without /utf-8? Those are the files that need to be looked at. If the failure is in comments, then it can be ignored/changed. But if it's in string literals, then they need to be looked at to verify that they are utf-8 or else it'll be a problem.

kjk:
I disabled 4819 warning instead.

PR #3410, opened 2023-03-31
Move Document colors from AppColors to theme.

avorobey:
For now we maintain compatibility with gPrefs->invertColors, gPrefs->fixedPageUI.textColor etc. by always going through getDocumentColors(), and using those prefs for the light (default) theme only.

This diff should be no-op for the default theme, and starts looking promising on switching to dark theme. May need work on invalidating cached prerendered pages in non-open tabs.


kjk:
Thanks!

PR #3418, opened 2023-04-04
Enable crash reporting in arm64

Biswa96:


Biswa96:
Here is a sample crash report in arm64 https://pastebin.com/zpytcc7T (temporary link).

kjk:
Thanks!

PR #3421, opened 2023-04-06
Complete move from AppColors.cpp to Theme.cpp.

avorobey:
- main window colors are now thematized, including canvas around doc
- AppColors.{h,cpp} not deleted, in a separate later commit. No callers remain
- these AppColor values are collapsed into "main window background": NoDocBg, LogoBg, MainWindowBg, AboutBg. They were in practice almost always the same
- the use of gSysColors is obsoleted (were consuled in NoDocBg). The use of #ABOUT_USE_LESS_COLORS is dropped.
- on the other hand, mainWindowBackground in gprefs is consulted when in the light theme, for backward compatibility.


kjk:
Thanks!

PR #3423, opened 2023-04-06
Thematize menus, both main and popup.

avorobey:
A theme now indicates whether its controls should be standard Windows colors, or we tinker as much as possible to adapt them to the theme. The default light theme leaves controls alone. The dark theme should try to adapt menus, edit boxes, static controls, toolbars, etc.


kjk:
Thanks!

PR #3428, opened 2023-04-08
Thematize toolbar

avorobey:
Renamed captionBackgroundColor to controlBackgroundColor, the color to be used on the background of all controls, captions, menus etc. Pure white in the default theme (but also code should preferably not even use it in the default theme, honoring theme->colorizeControls == false).

No-op on the default theme. Dark/darker themes look OKish... they need some tweaking on the bitmaps which look a bit wonky. I don't fully understand why the bitmaps look much better when switched to the About window, will look into it separately.

### DIFF
--- a/src/Caption.cpp
+++ b/src/Caption.cpp
@@ -165,7 +165,7 @@ void CaptionInfo::UpdateColors(bool activeWindow) {
                                                                   : GetSysColor(COLOR_INACTIVECAPTIONTEXT);
     }
     if (gGlobalPrefs->useTabs) {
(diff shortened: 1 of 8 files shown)


avorobey:
Feel free to delete a commented block at the beginning of  WndProcToolbar() in Toolbar.cpp, I forgot to remove it before committing. It's no longer needed.

kjk:
Thanks!

kjk:
I've created https://github.com/sumatrapdfreader/sumatrapdf/issues/3429 to track bugs / ideas / remaining task for theming

PR #3436, opened 2023-04-09
When possible use pdf_has_unsaved_changes(), which resets on save.

avorobey:
If the file was repaired on load (relatively rare), switch to inferior tracking of our own.


kjk:
Looks good but breaks dll build:

```
         EngineMupdf.obj : error LNK2001: unresolved external symbol _pdf_was_repaired [D:\a\sumatrapdf\sumatrapdf\vs2022\PdfFilter.vcxproj]
```

For dll build you need to add `pdf_was_repaired` to `libmupdf.def`

PR #3437, opened 2023-04-09
When possible use pdf_has_unsaved_changes(), which resets on save

### LINKED ISSUE #3417: Looks like a highlight bug.
1. left click and choose some text
2. right click > `Create Annotation From Selection` > `Highlight`
3. `Save changes to existing PDF`, and close this pop window
4. press hot key : `Ctrl + Shift + L` (Full Screen)
5. press `Esc` and then can see texts in red color: You have unsafed annotations.

avorobey:
If the file was repaired on load (relatively rare), switch to inferior tracking of our own. Fixes #3417.


kjk:
Thanks!

PR #3534, opened 2023-06-12
fix home tab's initial state is highlighted

huangshaohuai:
UseTabs = true 
NoHomeTab = false
open a pdf file -> close -> open, home tab state is always highlighted


kjk:
Thanks!

PR #3542, opened 2023-06-20
Add "QuietUninstallString" to registry

kriswilk:
**CHANGE:**

This PR adds the registry key `QuietUninstallString`, providing the command line required to silently uninstall the application.

**MOTIVATION:**

1) Although `QuietUninstallString` isn't mandatory, it is customary for uninstallers that offer a silent option.

2) The [WinGet](https://github.com/microsoft/winget-cli) package manager attempts to install/uninstall applications silently by default. The [SumatraPDF.SumatraPDF](https://github.com/microsoft/winget-pkgs/tree/master/manifests/s/SumatraPDF/SumatraPDF/3.4.6) package installs silently, but runs the uninstaller interactively. This occurs because WinGet falls back to `UninstallString` if `QuietUninstallString` is missing.


kjk:
Thanks!

PR #3552, opened 2023-06-25
Dark Mode with toolbar icon

tusharad:
On the toolbar I have added the darkmode icon (half moon). Which changes the text color to #ffffff and background-color to #000000. After changing the color, It calls RerenderEverything() which re-renders everything and changes color in real-time.


kjk:
Thanks but this is not how dark mode is supposed to work. See Theme.cpp

Also, I probably don't want to add another toolbar button by default and we don't have a way to customize the toolbar.

PR #3642, opened 2023-09-09
Support reversed pages

suacrbah:


kjk:
For backwards compatibility we cannot rename `CmdToggleContinuousView` - it can be used by name in settings file when configuring key bindings.

What's the motivation for adding those modes? I.e. when is it useful to display pages in reverse order?

suacrbah:
Sorry, I will recall the pull request. After consideration I think these modes may not be a very helpful function for the general user. These are like some tweaks only for me and at most a few more people.


To clarify the use case, I'm reading some ancient Chinese books whose pages and lines are in right-to-left style, this made reversing and horizontal continuity extremely useful for me. But I'm not sure if this is what most people wanted.


Thank you for reaching out. I will get the PR down immediately.




Suacrbah
***@***.***



&nbsp;




------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "sumatrapdfreader/sumatrapdf"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2023年9月12日(星期二) 凌晨5:39
***@***.***&gt;;
***@***.******@***.***&gt;;
主题:&nbsp;Re: [sumatrapdfreader/sumatrapdf] Support reversed pages (PR #3642)





 
For backwards compatibility we cannot rename CmdToggleContinuousView - it can be used by name in settings file when configuring key bindings.
 
What's the motivation for adding those modes? I.e. when is it useful to display pages in reverse order?

GitHubRulesOK:
@suacrbah 
Hebrew Arabic reverse Chinese or Japanese should all be readable right to left page order using "Manga Mode" combined with Facing / Book. So possible using existing commands. Just not always obvious as only shown on menu for say comic books, but usable for eBooks/PDF too

suacrbah:
@GitHubRulesOK 
I've seen "Manga Mode" during my modifying of the code and somehow find it unintuitive personally, but I agree with what you said, maybe I should reuse that and create another PR.

Issue #2816, opened 2022-07-22
Deleting current annotation will automatically jump to the first annotation


jiaan:
I have a pdf with many annotations such as ink and highlight, and I wish to remove certain annotations in a page. 
After opening the annotation editor and delete one of the annotations, the annotation editor automatically select the first annotation in the file, directing me to a different page in the pdf. I have to manually scroll down a long list of annotations to find the next one. This is very annoying. It is possible to fix this?

inkandvijay:
I am facing the same issue.

kjk:
I believe this is already fixed in latest pre-release https://www.sumatrapdfreader.org/prerelease

GitHubRulesOK:
@kjk not so sure using a file with annots on several pages in fixed page mode deleting one on 7 jumped me to page 1
![image](https://user-images.githubusercontent.com/12754913/225766953-7f75bfe7-f838-4450-98c9-dde20de1c07e.png)

kjk:
hmm, which method of deleting?

GitHubRulesOK:
dialog button so focus is in editor

kerryland:
I've just run into this too in 3.5.15481. Quite frustrating in a long document.

When I click "Delete annotation" in the screenshot below I do not expect any navigation to occur. Currently it moves me to the first annotation in the list.

![image](https://github.com/sumatrapdfreader/sumatrapdf/assets/398113/44cfe0d7-2bbe-470f-94c1-06176cd4b7f0)

PR #3654, opened 2023-09-14
Deletion of an annotation should not adjust PDF view

kerryland:
This is a fix for https://github.com/sumatrapdfreader/sumatrapdf/issues/2816 -- it was surprisingly easy!


kjk:
Thanks!

PR #3657, opened 2023-09-15
ctrl-dblclick on annotations is shortcut to edit

kerryland:
Users can now `ctrl-doubleclick` on an existing annotation and they will be taken to annotation editing. This is the same as the `"Edit Annotation"` context menu, but **much** more convenient to use when working with many annotations.

I'm less confident of this patch as it required more than one line of code. Also, I don't know how to communicate the _existence_ of this `ctrl-doubleclick` behaviour to the user. 

This partially fixes one of my gripes in bug#3638, but I will *try* and address the "needs a visual indicator to highlight the specific text" complaint in another pull request.

### DIFF
--- a/src/Canvas.cpp
+++ b/src/Canvas.cpp
@@ -445,7 +445,7 @@ static void SetObjectUnderMouse(MainWindow* win, int x, int y) {
         }
     }
 }
-
+// For selection/dragging
 static void OnMouseLeftButtonDown(MainWindow* win, int x, int y, WPARAM key) {
     // lf("Left button clicked on %d %d", x, y);
     if (IsRightDragging(win)) {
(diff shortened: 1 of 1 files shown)


GitHubRulesOK:
hmm untested but double click (unshifted) WAS invoke Synctex, more recently sullied by Double TAP as escape fulllview , so starting to wonder if there is a conflict in usages brewing?

kerryland:
I'm happy to tweak things some more if you can provide some guidance

kjk:
Thanks, I've used your code and tweaked it a bit in https://github.com/sumatrapdfreader/sumatrapdf/commit/5c63c533403e140ca8dc1051ef629a8166ad8ad2

@GitHubRulesOK the way it's written it shouldn't affect other use cases like selecting a word or synctex

Regarding exiting full screen: I'm open to other ideas for how to exit fullscreen mode with mouse / touch keyboard only. Double click was easy to implement.

GitHubRulesOK:
@kjk 
I agree this case is different and I am off topic, so regards double tap interfeers with historic double click my suggestion was perhaps add a delay between click interval so 2 fast taps is as was but 2 slower taps is exit screen mode, not a good solution 
(non intuitive)  but reduces the overlapping actions to more useful to majority who will constantly invoke double click.

Issue #1996, opened 2021-06-20
Indicate currently edited annotation on the page


kjk:
It's hard to tell which annotation is being edited. Indicate currently selected annotation in the editor on the page by e.g. drawing a focus ring around it.

GitHubRulesOK:
@kjk 
Apart from the fact is a horrible thick black slap in the face, this seems to have been addressed ?

kjk:
yep, thanks for spotting this

Issue #3638, opened 2023-09-07
Comments are truncated


kerryland:
As per issue #3086, "long" comments are truncated, which makes reading them very difficult (and reading comments is ever so useful!)

![image](https://github.com/sumatrapdfreader/sumatrapdf/assets/398113/e3581469-fa05-4d75-8581-41bf938bceef)

I discovered this after trying the pre-release build, because 3.4.6 had the flickering long comment bug. Truncating the text may stop the flickering, but it's not really a solution :-)

kerryland:
Here is a [sample pdf](https://github.com/sumatrapdfreader/sumatrapdf/files/12553034/sample.pdf)

kjk:
How does it look in other PDF readers?

We truncate comments because they are shown using a tooltip control and it doesn't do well with large text: it flickers, it's very slow and if we don't truncate at all, it might create a very large tooltip.

GitHubRulesOK:
@kjk 
I know we had problems with larger tooltips and Edge seem to also "truncate" tooltips to a manageable area, however Acrobat tends to use scroll boxes in line (in screen modal pop-up) and comments sidebars
![image](https://github.com/sumatrapdfreader/sumatrapdf/assets/12754913/9b55d380-bf20-4d17-868e-2c396f51a926)

kjk:
I get the need for it but doing what Acrobat does would be a significant change.

GitHubRulesOK:
I wonder the proposed free text enhancement allows such as this !! for better or worse 
Would it be posible to make that modal as popup ?
![image](https://github.com/sumatrapdfreader/sumatrapdf/assets/12754913/9b24381d-1ee2-4df9-b8d0-e669ebe86f6c)

kerryland:
> How does it look in other PDF readers?
> 
> We truncate comments because they are shown using a tooltip control and it doesn't do well with large text: it flickers, it's very slow and if we don't truncate at all, it might create a very large tooltip.

What was wrong with the implementation in Sumatra 3.3? It seemed to work OK with very large comments, and did not flicker.

That said, I have just discovered the "Annotation Editor", and I think this issue could be nicely solved with minor improvements in the integration between the Annotation Editor and the PDF viewer. Something like this:

- Double-clicking on a comment in the viewer should be the same as "Select Annotation in Editor". This would make it very easy to work through comments made on my document. Hurrah!
[...]

Issue #3590, opened 2023-07-27
PDF jumps up after deleting the last annotation


wskaa:
Jumps up = jumps to the page containing the first annotation (which in my case is the uppermost annotation in the document). This only happens when the annotations editor is open. 

A potential solution I considered is shortcutting a command for closing the annotation editor (but this command doesn't exist).

In case this needs to be said: it is a problem because it disrupts the annotating process by moving the user away from the page where they last wanted to annotate. In cases where the pdf file is large and is being read top to bottom (as in my case - I'm reading a long book), the problem increases as reading goes on (because you jump more pages to the first annotation).

GitHubRulesOK:
probably will close as duplicate
If I can find the one I seem to recollect raised in the past
Also may have been fixed in a pending fork but have not tested if that is its current behaviour?

GitHubRulesOK:
cant find the duplicate yet but is corrected in this fork
https://github.com/youmjb/SumatraPDF-mod

kjk:
I believe this is fixed in latest pre-release.

Specifically my test:
* create 2 annotations on 2 separate pages
* delete second annotation in annotations editor

In 3.4.6 this would select the first annotation in listbox of annotations and automatically jump to the page with first annotation.

Now it selects but doesn't auto jumps to the page.

PR #3661, opened 2023-09-16
The annotation currently being edited is now clearly identifiable

kerryland:
When editing annotations it's now each to tell which piece of text is related to the comment.

A couple of notes:
- My change to `EditAnnotations::StartEditAnnotations` was simply because I was offended by the repetition of calls to `SetColor` :-)
- `EditAnnotations::SelectAnnotationInEditWindow` has a new call to `RepaintAsync`, which smells a bit to me. There might be a better way to achieve the same goal, but I don't know it.
- I apologise if this pull request includes some stupid merges from master, my git skills are a bit rusty

This change fixes another issue I raised in [issue 3638
](https://github.com/sumatrapdfreader/sumatrapdf/issues/3638)

image


kerryland:
This fixes https://github.com/sumatrapdfreader/sumatrapdf/issues/1996

kerryland:
I've realised, after reading https://github.com/sumatrapdfreader/sumatrapdf/issues/3590, that this change makes deleting an annotation annoying, so I will fix that first.

src/EditAnnotations.cpp
```diff
@@ -1535,26 +1548,40 @@ Annotation* EngineMupdfCreateAnnotation(EngineBase* engine, AnnotationType typ,
 
     pdf_update_annot(ctx, annot);
     auto res = MakeAnnotationPdf(epdf, annot, pageNo);
+
+    std::optional <PdfColor> col;
```
kjk, line 1552: We don't use std:: if we can avoid it. It causes terrible bloat of executables.

Set to `ColorUnset` instead.

src/WindowTab.h
```diff
@@ -64,6 +64,13 @@ struct WindowTab {
     bool IsDocLoaded() const;
     void MoveDocBy(int dx, int dy) const;
     void ToggleZoom() const;
+
+    struct {
+        bool show = false;
+        RectF rect;
+        int page = 0;
+        bool scrolled = false; // only automatically scroll once
+    } currentEditAnnotationMark;
```
kjk, line 73: To be consistent with our naming scheme, it should be `CurrentEditAnnotationMark`

PR #3662, opened 2023-09-16
fixed bugs about onlyTabs Commmad

DudeRenderer:
- Launching onlyTabs command, entry tab not working in multiple windows.
- Selection Item in onlyTabs command is not correct in multiple windows with the same name tab.


kjk:
Thanks!

Issue #1688, opened 2020-07-14
Annotation GoTo


GitHubRulesOK:
Currently selecting an annotation will only goto the top of its page

It would be useful when there are many similar annotations on a page or zoomed in to be able to jump directly to the target xy location.

A bonus of this ability would be that users could apply annotation as visual "bookmarks or comment/reminder notes" which if saved OR NOT (for temporary useage) could close out https://github.com/sumatrapdfreader/sumatrapdf/issues/545 

One example usage is to select a piece of text, Press A to highlight it, go elswhere and navigate back exactly to that passage using the show annotations dialog

GitHubRulesOK:
ability has been met

PR #3663, opened 2023-09-16
The annotation currently being edited is now clearly identifiable

kerryland:
When editing annotations it's now each to tell which piece of text is related to the comment:

image

A couple of notes:
- My change to `EditAnnotations::StartEditAnnotations` was simply because I was offended by the repetition of calls to `SetColor` :-)
- `EditAnnotations::SelectAnnotationInEditWindow` has a new call to `RepaintAsync`, which smells a bit to me. There might be a better way to achieve the same goal, but I don't know it.

Solves most of my problem raised in [issue 3638](https://github.com/sumatrapdfreader/sumatrapdf/issues/3638), somebody else's in [issue 3590](https://github.com/sumatrapdfreader/sumatrapdf/issues/3590), and [issue 1996](https://github.com/sumatrapdfreader/sumatrapdf/issues/1996)

PS: Sorry about the additional commits in this pull request that are actually just dumb merges from master, my git skills are rusty, but I don't think it matters.


kerryland:
I think this also fixes https://github.com/sumatrapdfreader/sumatrapdf/issues/1688

src/EditAnnotations.cpp
```diff
@@ -1535,26 +1549,40 @@ Annotation* EngineMupdfCreateAnnotation(EngineBase* engine, AnnotationType typ,
 
     pdf_update_annot(ctx, annot);
     auto res = MakeAnnotationPdf(epdf, annot, pageNo);
+
+    std::optional <PdfColor> col;
```
kjk, line 1553: We don't use std::, please use ColorUnset instead

kerryland: Fixed.

kjk:
Thanks!

kerryland:
FYI there are some minor bugs that I need to fix:
1. Navigation via the `ctrl-dblclick` function doesn't update the highlighted annotation immediately
2. Navigation to `CmdEditAnnotations` via a keyboard shortcut doesn't update the highlighted annotation

PR #3665, opened 2023-09-17
"Select Annotation in Editor" didn't work when triggered via a shortcut key.

kerryland:
Here are a few small fixes for https://github.com/sumatrapdfreader/sumatrapdf/pull/3663

- I discovered an existing bug: *"Select Annotation in Editor"* didn't work when triggered via a shortcut (accelerator) key.
- I added a default accelerator key, `E`, for *"Select Annotation in Editor"*
- ctrl-double-click on an annotation didn't highlight the annotation
- I removed some logic that was adding `null` annotations to the annotation list, even though it didn't seem to cause a problem

### DIFF
--- a/src/Accelerators.cpp
+++ b/src/Accelerators.cpp
@@ -201,6 +201,9 @@ ACCEL gBuiltInAccelerators[] = {
     {0, 'u', CmdCreateAnnotUnderline},
     {0, 'U', CmdCreateAnnotUnderline},
 
+    {0, 'e', CmdEditAnnotations},
+    {0, 'E', CmdEditAnnotations},
+
     {0, 'i', CmdInvertColors},
     {0, 'I', CmdTogglePageInfo},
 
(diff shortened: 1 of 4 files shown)


kjk:
Thanks!

Not sure if I keep the 'e' shortcut, use can always configure that in settings.

kerryland:
> Not sure if I keep the 'e' shortcut, use can always configure that in settings.

Up to you, but I don't see the harm in providing a default (the user can always change it if they don't like it for some reason).  
I would be happy to remove the uppercase 'E' shortcut though, I just added that one for consistency with the others.

kjk:
My thinking is that single key is easy to hit accidentally so they should be reserved for things that are done frequently.

My current thinking is:
* double-click to select an annotation without opening edit annotations window
* ctrl + double-click to select and open annotations window

kerryland:
That's fine by me. Do you think double-click should update the annotations window, if the window is open?

kjk:
Yes, that makes sense.

Issue #3551, opened 2023-06-24
Saving annotations corrupts PDF file


actimem:
I recently updated to the latest version 3.4.6 64-bit.  I am using Windows 10.

When I open any PDF file, if I annotate it, I can see the "You have unsaved annotations" message at the top as expected.

Then when I save the annotations with Ctrl + Shift + S, I can see the "Saved annotations to ..." message, and the annotations are saved to file.

HOWEVER:
1. I can still see the "You have unsaved annotations" message at the top, i.e. it doesn't disappear.
2. If I attempt to close the file, the "Unsaved annotations" popup warning shows.  
3. If I then select "Save to existing PDF", it always CORRUPTS the file

I have also noticed that when Ctrl + Shift + S, sometimes it also corrupts the file, but I just noticed this like once out of 10 tries.

zengpengkindle:
i hava the same problem
![image](https://github.com/sumatrapdfreader/sumatrapdf/assets/16894507/2f751629-463c-47ce-9df9-dac7537e6cc0)

kerryland:
Does anybody have an example file?

kjk:
Could you provide more information:

* Does it still happen with latest pre-release https://www.sumatrapdfreader.org/prerelease ?
* how big is the PDF file?
* is the file on a network drive (e.g. One Drive, DropBox, network storage) ?

I tried a couple of times by adding / moving a bunch of annotations in a small PDF file and couldn't reproduce

Related to #3470 3470

kjk:
Also, can you provide a sample file on which it always happens?

Some PDF files are malformed. Our PDF engine (mupdf) tries to fix malformed files but it's possible the corruption only happens on such malformed files.

actimem:
I can reproduce 100% of the time.

I just downloaded this book book from gutenberg.org, and then converted to PDF using Calibre version 6.17.0.  Note this happens with most of my PDFs, many that haven't been converted by Calibre.
https://mega.nz/file/LEcgDaQL#fNQ0yNtYKTSdR6d6N4JLtRrr9YiLhD92RcxLuDqLaFU

1. On page 4, I select "Emma Woodhouse, handsome, clever, and rich", and press "a" to annotate.
2. I then see the message "You have unsaved annotations" at the top.
3. I then press (Ctrl + Shift + S) to save the annotation for the same file.
[...]

PR #3678, opened 2023-09-18
Improvements to save

kerryland:
This is the first step towards fixing https://github.com/sumatrapdfreader/sumatrapdf/issues/3551. All I have done is refactored out a lot of duplicated code that helps with readability, and gives one "save" method that can be updated with the reloading etc code.

I don't think I'm the person to add "the reloading etc code" -- there are an awful lot of "if, buts, and maybes" in these methods.

**EDIT:** I added a cheap fix. I don't really like it, ~but it seems to work~. I would prefer all reloading to exist within the new `SaveDocument` method.

### DIFF
--- a/src/EditAnnotations.cpp
+++ b/src/EditAnnotations.cpp
@@ -352,29 +352,11 @@ static void ButtonSaveToNewFileHandler(EditAnnotationsWindow* ew) {
 
 static void ButtonSaveToCurrentPDFHandler(EditAnnotationsWindow* ew) {
     WindowTab* tab = ew->tab;
-    EngineMupdf* engine = GetEngineMupdf(ew);
-    const char* path = engine->FilePath();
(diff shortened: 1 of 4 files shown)


kerryland (commented):
This doesn't actually work.

kjk:
Thanks, I did something similar based on this.

The reason your change didn't fix the corruption:

```
    ReloadDocument(tab->win, false);
    CloseAndDeleteEditAnnotationsWindow(tab->editAnnotsWindow);
    tab->editAnnotsWindow = nullptr;
```

`ReloadDocument()` won't do anything if `tab->editAnnotsWindow` is non-null, so it has to be called after `tab->editAnnotsWindow = nullptr;`.

This is because `EditAnnotationsWindow` holds state that would become invalid after reload.

PR #3684, opened 2023-09-19
Annotation editor switches tabs

kerryland:
Here are some minor quality-of-life improvements to the Annotation Editor I discovered while trying to work with multiple annotation editors being open at the same time.

1. When I click on an Annotation Editor, I now see the PDF content that the Annotations are related to. Previously I had to manually click on the tab, which got pretty tedious.
2. The Annotation Editor Title now includes the name of the associated PDF, which makes it easier to tell which window is which

### DIFF
--- a/src/EditAnnotations.cpp
+++ b/src/EditAnnotations.cpp
@@ -1384,8 +1384,11 @@ void StartEditAnnotations(WindowTab* tab, Vec<Annotation*>& annots) {
     args.icon = LoadIconW(h, iconName);
     // mainWindow->isDialog = true;
     args.bgColor = MkGray(0xee);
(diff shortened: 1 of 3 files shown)


kjk:
Thanks. I'll merge this after I merge the annots branch which refactors annotation handling.

As a reminder: we don't use `std::string`, that could be `str::JoinTemp()`.

kerryland:
> As a reminder: we don't use `std::string`, that could be `str::JoinTemp()`.

Sure. I don't have much C++ experience, and there seem to be so many options!

kjk:
Thanks. Couldn't merge due to conflicts with my other changes but applied those manually.

Issue #3688, opened 2023-09-20
Return to start page after you have finished searching


kerryland:
After I have finished searching for text I would like to be able to easily return to the page I was on when I initiated the search.

To reproduce:
- Imagine you are a famous author
- Open the new novel you are writing (eg: https://github.com/sumatrapdfreader/sumatrapdf/files/12641281/bug3551-Emma.-.Jane.Austen.pdf)
- Go to page 10
- Wonder "what was it I said about the umbrella"
- Search for "Umbrella"
- Read the fascinating words
- Press **ESC** in the search box.
- Be surprised, because nothing happened.

I expected to return to where-ever I was when I decided to read about umbrellas, instead I have try and remember where that was, and navigate back manually.

I'm aware I could do a rather more cumbersome version of this with "Favorites", but this behaviour would be handy.

kjk:
You can use `Alt - Left` to go back to page 10. We add navigation points and allow navigating back and forward like in a browser.

kerryland:
> You can use `Alt - Left` to go back to page 10.

Ah. This is cool! It just needs a minor tweak though, because currently `Alt - Left` is not captured when you are in the 'Find' input field, which is where you are when you are searching for stuff :-)

tactilis:
@kerryland   FYI  https://www.sumatrapdfreader.org/docs/Keyboard-shortcuts


> It just needs a minor tweak though, because currently Alt - Left is not captured when you are in the 'Find' input field, which is where you are when you are searching for stuff 

Agreed. Browsers allow ALT Left/Right navigation regardless of where the focus is.

PR #3689, opened 2023-09-20
ESC in 'Find' dialog returns start page

kerryland:
I've spent most of today working through a rather large pdf, adding lots of comments, and doing lots of searches, so I've found an irritation, which I have scratched :-)

This is a probably-incomplete implementation of this issue: https://github.com/sumatrapdfreader/sumatrapdf/issues/3688

It's really intended to get some feedback on whether you all think it's a good idea or not.

PS: The configuration setting `EscToExit` is very surprising. Does *anybody* use it?

### DIFF
--- a/src/SearchAndDDE.cpp
+++ b/src/SearchAndDDE.cpp
@@ -67,6 +67,7 @@ bool NeedsFindUI(MainWindow* win) {
     }
     return true;
 }
+int initialPage = 0;
 
 void FindFirst(MainWindow* win) {
     if (win->AsChm()) {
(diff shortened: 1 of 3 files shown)


kjk:
Like I mentioned in the bug: is `Alt - Left` enough for this? Not sure if we add navigation point when search starts, but we could (with `DisplayModel::AddNavPoint()`).

My decision making here would be: are there other programs that behave that way or would SumatraPDF be the only one with this behavior?

I think navigating on clearing search box would be unexpected behavior and a departure from what we've been doing for 16 years.

As to `EscToExit` : we've added it because someone requested it.

GitHubRulesOK:
Esc to exit is used by many cases of integration
where SumatraPDF may be used as a "previewer" / pop-in pop-out
such cases were usually navigators such as Total Commander where click on file opens single window and esc closes the window

The issue with navigate back is normally an icon in Acrobat and the challenge is that may not work when using pop-up child links that open a fresh tab since such tab would need to both esc and return to parent tab, which is sometimes wanted and sometimes unwanted. SumatraPDF does not have an icon for goback.

GitHubRulesOK:
@kerryland 
oddly I was discussing the differences between Adobe desktop Adobe Browser and SumatraPDF recently see
https://stackoverflow.com/a/77138910/10802527

kerryland:
> Esc to exit is used by many cases of integration
> where SumatraPDF may be used as a "previewer" / pop-in pop-out

It makes sense in that context. Thank you both for your patience.

kjk:
I assume `Alt + Left` addresses tihs.

PR #3767, opened 2023-10-15
Update pt.txt

smarquespt:
@kjk I've also updated strings in apptranslator. Please ignore them or upload this file in there.

Thanks


kjk:
Thanks, no need to do PR for translations from apptranslator. They will eventually be updated.

PR #3921, opened 2023-11-28
Allow setting output filename in `-print-settings` from command line

### LINKED ISSUE #3917: Allow output file in `-printer-settings` from command line
I was reading `-print-settings <settings-list>` in the [docs](https://www.sumatrapdfreader.org/docs/Command-line-arguments) and was wondering if it is feasable to add `output=<filename>` in the advanced options.

I looked at the sources and a PR would be 10-20 lines of code for this feature if you would allow such addition.

The most obvious use-case is to use `Microsoft Print To PDF` printer with an output file to convert some PDFs to other PDFs for one reason or another.

wqweto:
Closes #3917

### DIFF
--- a/src/Print.cpp
+++ b/src/Print.cpp
@@ -101,6 +101,7 @@ void Printer::SetDevMode(DEVMODEW* dm) {
 
 Printer::~Printer() {
     str::Free(name);
+    str::Free(output);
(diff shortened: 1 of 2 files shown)


kjk:
Thanks!

PR #4280, opened 2024-05-28
Add new search features

prettycrewcutyulia:
Adding quick search functionality to Wikipedia and Google Scholar, which would be extremely useful for students and scientists


kjk:
Thanks!

PR #4282, opened 2024-05-28
Update ExternalViewers.cpp

mkim0710:
Registry value for Foxit PDF Reader 12.1.3.15356 (The last version with Add Bookmark function without bugs in single-key accelerator)


kjk:
Thanks!

PR #4294, opened 2024-06-01
Number of recently used files is configurable in expert mode

stassie-lzrv:
It will be useful to have more or less than 10 recently used files displayed in history

### DIFF
--- a/do/settings_def.go
+++ b/do/settings_def.go
@@ -319,6 +319,8 @@ var (
 			"if true, we use Windows system colors for background/text color. Over-rides other settings").setExpert(),
 		mkField("UseTabs", Bool, true,
 			"if true, documents are opened in tabs instead of new windows").setVersion("3.0"),
+		mkField("MaxRecentlyUsedFiles", Int, 10,
+			"number of most recently used files that will be displayed in the menu").setExpert(),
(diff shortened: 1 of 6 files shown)


kjk:
How many? I'm not opposed to increasing this number but adding a config option for it is too much.

Other ways to improve this:

1. Improve Command Palette

Currently `Ctrl + K` Command Palette shows recent files at the top followed by commands. That is a better way of selecting a file from history but could be improved.
Add "Show just file histry" mode, triggered by e.g. "f:" prefix. Make it a command so that can it can be bound to a keyboard shortcut. Add help text in Command Palette or a button at the top to switch between modes to make it obvious to people those are possibilities.

2. Make thumbnails scrollable so that user can scroll the whole history.

3. Add typedown search to thumbnails area i.e. a text box and when you type partial file name, we only show thumbnails matching that partial name.

GitHubRulesOK:
@kjk just a comment
some apps will on click file  open icon offer a dropdown of "recent's" (so in SumatraPDF via explorer I see 14 then a down arrow) that in some cases has a "showmax" setting so last 5  or  last 10 but most users opt for whatever is maximal !
Unsure if a similar ability may be taken with  an Alt "file open" accelerator as its logical there

PR #4381, opened 2024-07-18
Initial clang-cl build

R-Goc:
An initial clang-cl build. For purposes of building with CI it is not ready as an architecture is set manually as clang notices some library functions have a dependency on specific features, and doesn't build without them, notably sse4.1 and aes. Both are old, but setting a specific feature set could be breaking and suboptimal, so needs further discussion. Currently clang-cl build is with warnings disabled. 
Passing qsort _stricmp directly is UB, though the way I resolved it currently probably isn't the prettiest. 
Sadly can't get it to build without exceptions on clang.


R-Goc:
the commits should probably be squashed

kjk:
Please:
1. delete `Directory.build.props` it points to your personal setup
2. `.gitignore` change is empty so either you didn't make the change you wanted or it should not be changed
3. squash the commits into a single commit

Other than that looks good.

kjk:
Thanks!

PR #4382, opened 2024-07-18
Ctrl+Tab in the most recent order

DaPa:
This change implements the https://github.com/sumatrapdfreader/sumatrapdf/discussions/2750 request via an advanced option CtrlTabLastViewed (default false). When CtrlTabLastViewed=true, then Ctrl+Tab behaves as in Firefox, going back to the most recent tab.
If Ctrl+Tab is pressed but Ctrl is not yet released then "CmdCommandPaletteOnlyTabs" is triggered so that the user can jump to another Tab, similar to Firefox's behaviour:
SumatraPDF_oUhoicHX2r

### DIFF
--- a/do/settings_gen_def.go
+++ b/do/settings_gen_def.go
@@ -372,6 +372,8 @@ var (
 			"3 is fullscreen, 4 is minimized"),
 		mkCompactStruct("WindowPos", windowPos,
 			"default position (can be on any monitor)").setStructName("Rect").setDoc("default position (x, y) and size (width, height) of the window"),
+		mkField("CtrlTabLastViewed", Bool, false,
+			"if true, Ctrl+Tab cycles through tabs in recently used order").setVersion("3.6"),
 
 		// file history and favorites
 		mkArray("FileStates", fileSettings,
--- a/src/CommandPalette.cpp
+++ b/src/CommandPalette.cpp
@@ -447,6 +447,9 @@ void CommandPaletteWnd::CollectStrings(MainWindow* mainWin) {
 
 static CommandPaletteWnd* gCommandPaletteWnd = nullptr;
 static HWND gHwndToActivateOnClose = nullptr;
+// true WHEN gGlobalPrefs->ctrlTabLastViewed AND gCtrlTabPressed: convert "#" to "@"
+static bool gCtrlTabLastViewed = false;
+
 
 void SafeDeleteCommandPaletteWnd() {
     if (!gCommandPaletteWnd) {
@@ -480,7 +483,7 @@ LRESULT CommandPaletteWnd::WndProc(HWND hwnd, UINT msg, WPARAM wparam, LPARAM lp
 }
 
 bool CommandPaletteWnd::PreTranslateMessage(MSG& msg) {
-    if (msg.message == WM_KEYDOWN) {
+    if (!gCtrlTabLastViewed && msg.message == WM_KEYDOWN) {
         int dir = 0;
         if (msg.wParam == VK_ESCAPE) {
             ScheduleDelete();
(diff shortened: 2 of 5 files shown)


kjk:
Thanks. I'm closing this because I've implemented something similar in https://github.com/sumatrapdfreader/sumatrapdf/commit/41af62c16e0795dc4b97a6d4210b6f7382a4af7d

There's no setting. The new default for `Ctrl + Tab` is `CmdSmartTabSwitch` which shows command palette in smart tab selection mode.

One improvement in smart tab selection mode is: if you release Ctrl right away, command palette stays and you can do name filtering and use up / down / enter / escape.

If you keep pressing `Ctrl + Tab` or `Ctrl + Shift + Tab`, it'll select the tab when releasing Ctrl.

Didn't implement recently selected order. It seems you've only been tracking the last selection (and not the full history).

Could add tracking of last tab and set the initial selection on that (currently selection is on current tab). Wouldn't be as smooth anyway because to select would have to release Ctrl and Enter.

Also in that mode could make single mouse click a selection (currently it's double-click).

The user can re-bind `Ctrl + Tab` to previous `CmdNextTab` and `Ctrl + Shift + Tab` to `CmdPrevTab` commands for previous behavior.

kjk:
Another option would be to add new commands. Either track the full history of tabs and add `CmdNextTabInHistory` and `CmdPrevTabInHistory`.

Or track just the last tab and `CmdSwitchToPrevTab` for switching between 2 tabs.

Or both.

Then users could bind them to whatever shortcut the want.

DaPa:
Thank you for your suggestions and comments! New implementation is indeed nicer as we now see better the name of the tabs which we want to switch (via the opened command pallete), but the quick switching to the last would still be welcome ;-) 

Firefox seems to keep track of only last 7 tabs:
- pressing and releasing Ctrl+Tab switches to latest tab
- pressing Ctrl+Tab and releasing Tab opens a shorter list of only 7 latest tabs
- from there the user has the option to continue Tabing to one of the latest 7 tabs and an option to open a list of all other tabs too:
firefox_gnN8gvLgqA

For SumatraPdf I think we need to track only the last-tab and if the user wants to select **another** last-tab then it's easier to open command pallete and select the wanted one: just an opinion based on my usage, but maybe others want different things like the full history tracking as you suggested... 
Thank you again for creating this excellent fast and lite Pdf tool!

kjk:
Another option: `CmdSelectTabFromHistory` bound by default to `Ctrl + H` (`h` for history) or `Ctrl + T` (`t` for tab.

It would display something similar to command palette:

image

It positions by default on last tab so just to switch to previous tab: `Ctrl + H`, `Enter`.

Also you can press `0` to `9` to switch to last n-th tab (displayed on the right).

This screenshot above is how I implemented it in [Edna](https://edna.arslexis.io/) (my note-taking app).

It could re-use command palette.

If you're feeling ambitious, you could give implementing one of those ideas a try. They are not terribly difficult to implement.

I think I like this version the most. It would allow for quick toggle between 2 documents but also allow navigating larger history.

usergitalberto:
Could someone explain to me how to go back to the old CTRL+TAB function that quickly switches to the next tab WITHOUT displaying an additional screen?
Thank you

DaPa:
> Could someone explain to me how to go back to the old CTRL+TAB function that quickly switches to the next tab WITHOUT displaying an additional screen? Thank you

Normally you should go to Settings -> Advanced Options and edit the Shortcuts section in the opened SumatraPDF-settings.txt:
```
Shortcuts [
	[
		Cmd = CmdNextTab
		Key = Ctrl + Tab
		Name = Previous CmdNextTab
	]
	[
		Cmd = CmdPrevTab
		Key = Ctrl + Shift + Tab
		Name = Previous CmdPrevTab
	]
]
```

image

I tried it, however it seems to not work! The new named shortcuts are seen in the command pallete fine:
image

But they just don't act... when pressing Ctrl+Tab it still triggers the new implementation, while Ctrl+Shift+Tab does nothing.
Tested with latest build:
image

Maybe @kjk can confirm and clarify?

kjk:
@DaPa thanks for noticing. Turns out `Tab` was never recognized in Shortcuts.

Fixed with https://github.com/sumatrapdfreader/sumatrapdf/commit/a698487418130e51aaddcfdbe1902d156e1eeea0

kjk:
@usergitalberto could you try again Ctrl + Tab again in latest pre-release and let me know if it works for you?

I changed it to behave like standard `Ctrl + Tab` / `Ctrl + Shift + Tab`.

The difference is that it also shows the list of tabs in a list and you can press `Space` to switch to "sticky" mode where the list won't go away when you release `Ctrl + Tab` so that you can filter the list by name.

usergitalberto:
Now it works perfectly.

(anyway some trouble with "search" function, but I see someone is still working to solve it!)
Thank you

kjk:
What do you refer to in "search function"? I don't know of anyone working on anything in SumatraPDF.

usergitalberto:
When I was testing CTRL+TAB (thank you again to reintroduce this function: it permits me to "compare" easly and quickly differences on 2 different bus similar documents) I used search/find (CTRL+F) tools and Sumatra crashed when inserting the second or third char..

Commit: Fix crash in search
Fix crash in search

kjk:
Yeah, I already fixed search crash in latest pre-release

PR #4391, opened 2024-07-22
update synctex version to 2023.4(TeXLive)

melven:
I had problems with synctex (some of the filenames got truncated).
So, I assume there is a synctex parser bug in the current (old) version of the synctex code in the ext/ directory.
This updates ext/synctex to a recent version with minor fixes for compiling on Windows with VS22.


kjk:
Did you test this code with your latex projects?

I don't use this functionality so I can't test it properly.

GitHubRulesOK:
Just an off the cuff observation that last update seemed to upset some reliability that was there before in handling synctex.gz so there were more "synctex" issues raised recently than before.

Part of that is the many many different editors have expected SumatraPDF to always behave as it did in the past, so testing would need to be fairly extensive in multiple editors.

melven:
> Did you test this code with your latex projects?
> 
> I don't use this functionality so I can't test it properly.

Setting this to output the file and line:
grafik

Before (v3.6 64bit Windows)  I got this:
grafik

With these changes, I get this correct path:
grafik


For PDF pages before this specific page, it works also with version v3.6.
That's why I assume that this is a parser bug.
I have a `<pdfname>.synctex.gz` file generated by latex in the folder of the pdf-file.

melven:
> Just an off the cuff observation that last synctex related update seemed to upset some reliability that was there before in handling synctex.gz so there were more "synctex" issues raised recently than before.
> 
> Part of that is the many many different editors have expected SumatraPDF to always behave as it did in the past, so testing would need to be fairly extensive in multiple editors.

My setup involves VsCode with the LaTeX-Workshop plugin which opens SumatraPDF standalone (described here: https://superuser.com/a/1785147)

I cannot really test many other setups.
As I have build a working SumatraPDF executable, I'll just use that for now in my daily work.
I would hope that the changes in synctex would fix problems and not introduce new unintended behavior but I'm not familiar with the synctex code, so I have no insight into that.

PS
I've added a `#define` in the synctex code to disable building the part that needs `gzwrite` (which failed in CI).

kjk:
Thanks. We'll see what happens. If people complain, can always revert.

PR #4582, opened 2024-10-12
feat(annotations): A pop up menu appended to annotation list box is added.

Void-Fleet:
A pop up menu appended to annotation list box is added. Now only 'delete selected annotation' function is available.
However the menu is activated via double clicking left button, so it's just a little weird.
image


kjk:
Three issues:

It doesn't add new functionality (there's `delete` button and you can press `Del` key)

Not going to add context menu for just one item.

Should be right-click. No one will figure out double click.

PR #4598, opened 2024-10-21
Open Explorer and select file

ajain-93:
Instead of just opening the directory, this selects the file. This helps if you have a folder with lots of similarly named file.

### DIFF
--- a/src/ExternalViewers.cpp
+++ b/src/ExternalViewers.cpp
@@ -44,7 +44,7 @@ static ExternalViewerInfo gExternalViewers[] = {
         CmdOpenWithExplorer,
         "*",
         "explorer.exe",
-        R"("%d")",
+        R"(/select,"%1")",
         nullptr,
         nullptr,
     },


GitHubRulesOK:
The command is for directory "CmdOpenWithExplorer  = Open Directory In Explorer" how does that relate to any single selection ? to jump to current file in a directory there are other commands

ajain-93:
It's basically an enhancement, opening the directory and selecting the file.

But, you are right. The `Show in Folder` option already accomplish this. I didn't see this option before, as I expected it to be after the seperator, not grouped with open, save etc.

GitHubRulesOK:
Will have to see what dev thinks 
Personally did not see the need for call the directory while focused on file ? but guess if the functionality is there, then no problem to add it for the use of ?

GitHubRulesOK:
To open the existing selected file **in new window** (can be dragged to existing tabs for 2 copies side by side)
It can be done using existing code without changes so in ExternalViewers call
```
	[
		CommandLine = "C:\program files\sumatrapdf\sumatrapdf.exe" -new-window -page %p "%1"
		Name = Same page in &New window
		Filter = *.*
		Key = N
	]
```
image

kjk:
thanks, makes sense

Issue #2186, opened 2021-10-13
Page size - ability to change the unit from CM to MM


kanlukasz:
Is it possible to change unit from `CM` to `MM`?
![image](https://user-images.githubusercontent.com/30685349/137086486-e29a3716-b933-4d16-aaca-b69115b1ce0d.png)

If not, is there any chance to add this option?

flywire:
> `m` - show cursor position in document coordinates

Be nice if this was customisable too

GitHubRulesOK:
@flywire 
oddly `m` uses mm (along with in & pt) the complaint is file details uses cm !

GitHubRulesOK:
@flywire press `m` several times it loops between native pt and alternative units

PR #4624, opened 2024-11-06
Correct keyboard shortcuts for show page number and invert colors

klsbecker:
- Updated the default keyboard shortcuts for the following commands:
  - 'Show current page number'
  - 'Show cursor position in document coordinates'
  - 'Invert colors in the document'

This update ensures that the shortcuts match the expected behavior, improving accessibility and user experience.

### DIFF
--- a/docs/md/Keyboard-shortcuts.md
+++ b/docs/md/Keyboard-shortcuts.md
@@ -88,8 +88,8 @@ You can [customize keyboard shortcuts](Customizing-keyboard-shortcuts.md). Also 
 - `F11`, `Ctrl + Shift + L`, `f` - Enter / Exit full screen mode
 - `ESC` - exit full screen or presentation mode
 - mouse double click - exit full screen or presentation mode
-- `i` - invert colors in the document
-- `Shift + i` - show current page number
+- `Shift + i` - invert colors in the document
+- `i` - show current page number
 - `m` - show cursor position in document coordinates
 - `F8` - show/hide toolbar
 - `F9` - show/hide menu


GitHubRulesOK:
Hmm the shifted `i` reversal for 3.6 is correct 

but the existing measure in 3.6 is still  
`m` - show cursor position in document coordinates

KlsBecker:
Hello,

I tested on Windows, and the **default** shortcuts are as described in the commit.

GitHubRulesOK:
it was normal `m` key 2 years ago
https://github.com/sumatrapdfreader/sumatrapdf/issues/2186#issuecomment-1350087438
 and as far as I know has always been normal unshifted m for **m**easure

Invert `I` did change for 3.6 but also current docs are **for current formal release 3.5.2** and there the 
normal `i` = **i**nfo
SHIFTED I = INVERT 
image

Thus the change will be needed later for the 3.6 and not current 3.5 release

kjk:
`Shift + i` change needs version info. it's `i` in 3.5 or earlier and `Shift + i` in 3.6+

klsbecker:
You're absolutely correct. I had Caps Lock on by accident. 🤡

Furthermore, for the new version, the shortcuts related to `Shift + i` and `i` must be switched to match the current behavior.

duberlyguarnizofuentesrivera (approved):

kjk:
Thanks, made similar change in https://github.com/sumatrapdfreader/sumatrapdf/commit/5c42f50c40a580bf291d15ea69ddcf3a3c4b09ab

PR #4663, opened 2024-11-26
Integrating ML mode with SumatraPDF

DharaniDJ:


GitHubRulesOK:
Hmm **I am not the developer** but using Python as a sideload is a giant maintenance overhead and for Mutool convert page to/from image there are already working internal routes (using shell and DDE) but again a big maintenance overhead that was **temporarily ??** dropped due to problems with multiple filetypes.

PR #4722, opened 2024-12-16
Move smooth scroll timers to a separate thread, instead of using `SetTimer`

goodtrailer:
Potentially addresses #4122

**Note.** To avoid confusion, I'm going to refer to middle-click scroll as "auto scroll." And I'm going to refer to smoothly animated/interpolated scrolling (enabled by the `SmoothScroll=true` config) as "interpolated scroll." In the code, "auto scroll" is referred to as `SMOOTHSCROLL`, while "interpolated scroll" is referred to as `smoothScroll`. This could be confusing, which is why I call them "auto scroll" and "interpolated scroll."

---

Discussion #4122 complains that "auto scroll" is too scratchy, and I noticed that this is probably because it uses `SetTimer()`. Also, "interpolated scroll" uses `SetTimer()` as well. An issue is that `SetTimer()` allows a minimum of 10ms/100fps (according to the [docs](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-settimer), but it seems the true minimum is closer to  16ms/60fps in my testing). Many displays run at much higher than 100Hz (e.g. mine runs at 144Hz), so this results in a very scratchy appearance.

A relatively non-intrusive solution is to replace the timer with a thread (see `ScrollTimerThread()`) which sends `WM_TIMER` messages to the canvas window. Since we use our own thread instead of `SetTimer()`, we can run at the monitor's refresh rate, instead of a capped 10ms/100fps. Note that using this method, not very much code needs changing. We still use `WM_TIMER` messages, which can be handled in the exact same way as before. The only difference is that instead of `SetTimer()`/`KillTimer()`, we have to call `SetEvent()`/`ResetEvent()` on an Event object which the thread is waiting on. Also, we have to account for the delta time (Δt) so that scrolling isn't faster/slower depending on refresh rate (i.e. for "auto scroll," a 144Hz monitor and a 60Hz monitor should have the same scroll speed).

---

Note that this is a draft PR because I just wanted to show a proof-of-concept that improves scrolling on high-refresh displays, in case it is helpful at all. It's not perfect, since `WM_TIMER` messages are [low-priority](https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-timer#remarks). But it's relatively simple and non-intrusive, which is why I wanted to bring it up.

### DIFF
--- a/src/Canvas.cpp
+++ b/src/Canvas.cpp
@@ -64,7 +64,7 @@ Kind kNotifAnnotation = "notifAnnotation";
 // Smooth scrolling factor. This is a value between 0 and 1.
 // Each step, we scroll the needed delta times this factor.
 // Therefore, a higher factor makes smooth scrolling faster.
-static const double gSmoothScrollingFactor = 0.2;
+static const double gSmoothScrollingFactor = 0.3;
 
 // these can be global, as the mouse wheel can't affect more than one window at once
 static int gDeltaPerLine = 0;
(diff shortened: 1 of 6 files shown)


kjk:
I'm not sure if I want to keep it but also not sure if I want to reject it.

I've moved this code to a branch for more testing at https://github.com/sumatrapdfreader/sumatrapdf/tree/smooth-scroll-thread and therefore I'll close this PR.

Issue #4030, opened 2024-01-17
Invert Colors not working on pre-release


Chivos:
**SumatraPDF version**
- Version pre-rel 3.6.15941 (and before)

**Describe the bug**
The Invert Colors function is not working anymore

**To Reproduce**
Try to invert colors using keybord shortcut or command palette

pulbitz:
I've been having this issue since the version that fixed it.

I'm using the Darker theme UI, with black set to full black and white set to low contrast.

Theme = Darker
FixedPageUI [
	TextColor = #000000
	BackgroundColor = #bbbbbb

When using the Invert Colors, if the Theme is Dark/Darker and not Light, the TextColor/BackgroundColor provided by the Theme will be used, not the Color value I set.
If TextColor and BackgroundColor are not the default and are custom set by the user, I would like them to respect the Color value regardless of the Theme.

kjk:
Please open a new issue for this because the bug reported here has been fixed. This is more specific issue.

PR #4832, opened 2025-03-07
Partial Revert of "fix color invert in default theme" (fixes #4073)

### LINKED ISSUE #4073: Since 3.6.15949, Invert Color feature in Dark/Darker theme does not respect custom colors.
**SumatraPDF version**
3.6.15949 pre-release or later

**Describe the bug**
I'm using the Darker theme UI, with black set to full black and white set to low contrast.

Theme = Darker
FixedPageUI [
TextColor = #000000
BackgroundColor = #bbbbbb

When using the Invert Colors, if the Theme is Dark/Darker and not Light, the TextColor/BackgroundColor provided by the Theme will be used, not the Color value I set.

If TextColor and BackgroundColor are not the default and are custom set by the user, I would like them to respect the Color value regardless of the Theme.

God-damnit-all:
Since https://github.com/sumatrapdfreader/sumatrapdf/commit/5adadc279436ed92195fe846ea0b3f8baad7fc59 as reported in #4073:
> **SumatraPDF version** 3.6.15949 pre-release or later
> 
> **Describe the bug** I'm using the Darker theme UI, with black set to full black and white set to low contrast.
> 
> Theme = Darker FixedPageUI [ TextColor = #000000 BackgroundColor = #bbbbbb
> 
> When using the Invert Colors, if the Theme is Dark/Darker and not Light, the TextColor/BackgroundColor provided by the Theme will be used, not the Color value I set.
> 
> If TextColor and BackgroundColor are not the default and are custom set by the user, I would like them to respect the Color value regardless of the Theme.

The code for `userDidChange` was erroneously removed, which was unnecessary to fix the bug reported in #4030.

### DIFF
--- a/src/Theme.cpp
+++ b/src/Theme.cpp
@@ -203,6 +203,13 @@ COLORREF ThemeDocumentColors(COLORREF& bg) {
         return text;
     }
 
+    // if user did change those colors in advanced settings, respect them
+    bool userDidChange = text != kColBlack || bg != kColWhite;
+    if (userDidChange) {
+        std::swap(text, bg);
+        return text;
+    }
+
     // default colors
     if (gCurrentTheme == gThemeLight) {
         std::swap(text, bg);


kjk:
thanks

PR #4892, opened 2025-04-14
Check image size when reading targa file

npt-1707:
**Description**
This PR throws an error when image width or height is 0 when reading a targa file in ext/libjpeg-turbo/rdtarga.c which was cloned from libjpeg-turbo but did not receive the patch applied in the original repository. The original issue was reported and fixed under [libjpeg-turbo/libjpeg-turbo@82923eb](https://github.com/libjpeg-turbo/libjpeg-turbo/commit/82923eb93a2eacf4a593e00e3e672bbb86a8a3a0).


kjk:
thanks, but the right fix is to upgrade the whole libjpeg-turbo to newer versions. As per `ext\versions.txt` we have a pretty old version 1.3.1

We should update to newer version, not cherry-pick checkins.

It just requires more work because it possibly requires updating `premake5.lua` and `premake5.files.lua` and re-generating the solution with `.\doit.bat -premake` (requires Go).

If upgrading to the latest version is too complicated, we can try updating to older tags, one by one.

PR #4948, opened 2025-05-31
Update Theme.cpp

nonverbalbychoice:
I like the "Darker" theme, but it is hard to read text at night.  
I made a new theme so i can see the text, when i am reading

### DIFF
--- a/src/Theme.cpp
+++ b/src/Theme.cpp
@@ -59,6 +59,14 @@ static const char* themesTxt = R"(Themes [
         LinkColor = #6B7280
         ColorizeControls = true
     ]
+    [
+        Name = Dark background Bright text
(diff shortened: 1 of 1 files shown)


nonverbalbychoice:
i'm wondering what's the easiest way to also invert colors. Like maybe we should be using "System Theme". Like, i have my Windows 11 theme set to dark. How do i integrate that? I'm not sure. Maybe somebody can help me :)

GitHubRulesOK:
@nonverbalbychoice 
There is only one "inversion" of foreground and background canvas colour range (what MuPDF calls "tinting") thus the `I` command is all that is currently available as shown in your image.
The default Dark theme is the blackest and yours is a bit brighter than that.
image

PR #4989, opened 2025-06-22
Improve theme for menu and toolbar for older OS

ozone10:
This PR fixes issues with theme is not applied when modifying custom toolbar buttons, and improve dark theme handling on older OS.

### DIFF
--- a/src/Menu.cpp
+++ b/src/Menu.cpp
@@ -23,6 +23,7 @@
 #include "DisplayModel.h"
 #include "FileHistory.h"
 #include "Theme.h"
+#include "DarkModeSubclass.h"
 #include "GlobalPrefs.h"
 #include "Annotation.h"
 #include "AppColors.h"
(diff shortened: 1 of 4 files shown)


kjk:
Thanks!

@ozone10 a question: I've added "solarized light" theme. I noticed that e.g. Find dialog (F8 to hide toolbar, Ctrl + F to invoke find) is not themed with solarized background color but it is for dark themes.

I assume that in the context of darkmodelib "dark" is just a marketing name and it's actually a theming library that will theme for any color set, including light colors like in solarized light?

ozone10:
It is because of `DarkMode::setDarkModeConfig(DarkMode::isColorDark(ThemeWindowControlBackgroundColor()) ? 1 : 3);`
For light themes it disables all theming for compatibility. you can replace value `3` with `0` to enable theming for light colors. I did not test it on unsupported OS so I've used classic style for light themes.

ozone10:
btw prerelease have version 3.6.17021 shouldn't it be 3.6.1**6**021

kjk:
For historical reasons I add 1000 to checkin count see getGitLinearVersionMust()

PR #4996, opened 2025-06-26
Make themes less confusing

user1823:
Previously, "Darker" had a brighter background color than "Dark", which is confusing.

### DIFF
--- a/src/Theme.cpp
+++ b/src/Theme.cpp
@@ -53,27 +53,19 @@ static const char* themesTxt = R"(Themes [
         ColorizeControls = true
     ]
     [
-        Name = Darker
-        TextColor = #c3c3c6
-        BackgroundColor = #2d2d30
-        ControlBackgroundColor = #2d2d30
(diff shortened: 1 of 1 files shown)


kjk:
Thanks.

Any interest in more radical change? The current dark themes are just me letting other people make decisions so they don't map to any known color schemes.

It would be better to add well-known color schemes like Solarized Dark (I've already added Light), One Half Dark / Light, Campbell, Tango Dark / Light. Those are from (the new) Windows Terminal. Other source of inspiration: Visual Studio Code.

Also, with darklibmode doing the theming we could adopt their colors i.e. more colors for more fine-grained control over colors of various elements.

user1823:
Thanks for the clarification. I submitted the PR hoping to simplify what I found confusing, but I understand you're aiming for a more comprehensive overhaul. Sorry, but I don’t have the bandwidth to take that on. Appreciate the feedback.

PR #5000, opened 2025-07-02
Update darkmodelib and enabled its support for light themes

ozone10:
This PR add full support for light themes similar to dark ones.


kjk:
Thanks!

Any interest in aligning SumatraPDF theming more with darklib i.e. mostly adding more color definitions to Sumatra's theme colors, passing them to darklib and updating the themes with those colors?

ozone10:
It is already aligning some of them
https://github.com/sumatrapdfreader/sumatrapdf/blob/master/src/Theme.cpp#L161-L169

for other default colors are used
https://github.com/sumatrapdfreader/sumatrapdf/blob/master/ext/darkmodelib/include/DarkModeSubclass.h#L278-L291

Neustradamus:
@ozone10: Thanks!

@kjk: A new build is possible?

GitHubRulesOK:
@Neustradamus the next prelease will include the above features
When that may be is down to other factors but will be the one **after** current latest pre-release: 3.6.17044

PR #5108, opened 2025-09-24
Fix single page print with modern dialog

### LINKED ISSUE #3453: Always printing the first page no matter which page I choose
This is a weird one.

Since I don't have a printer that flips the page automatically, I have to do it by hand. The way I do it is by sending the first page to print, put the paper back into the printer flipped and then I send the second page and somehow end up with the FIRST page on both sides.

I also have no preview of the page for some reason so I don't know which one it will spit out. ~Also tried printing using Microsoft Print to PDF, which does not produce this issue.~ Double checked, Print to PDF is printing the wrong page as well.

The printer is a Brother HL-1210W if that matters.
Using Windows 11 Pro 22H2 22621.1555.
sumatrapdf version is 3.4.6 x64.

Tried printing from Chrome, Word etc. and I could not reproduce the bug anywhere so I doubt it could be caused by the printer driver.

I've been using sumatrapdf for 10+ years and never had this issue in the past, maybe switching to Windows 11 is when it started?

### LINKED ISSUE #4235: Printing Issue
**Dear SumatraPDF**
I am writing to report a critical issue I encountered while using the Pages option in printing setting on PDF files.
**Version** _:- **pre-release build version 16081**

**The problem arises when attempting to print a specific page. Despite inputting the desired page numbers in the custom range, the printed output consistently reflects the page I am currently viewing.**

When printing, I go to pages in the print settings and enter any page number I want in the custom range (for example, I entered no. 2 in the page range), but if I use the custom range while I am on the first page (my current page is 1) and trying to print the 2nd page or any other custom page, the page I'm currently on, which is page 1, is printed. Even after entering my custom page number, that page is not printed; my current open page is printed. 
The custom page no, which I have given in the page custom range, should be printed. Now I need to go through each selected page that I need to print.
Screenshot 2024-05-12 103733

### LINKED ISSUE #4779: Printing only last page in Bookview not possible
**SumatraPDF version**
- Version 3.5.2, 64-bit, Windows 11 23H2

**Describe the bug**
Printing only the last page while in bookview (Ctrl-8) is printing the second last page.

**To Reproduce**
1. Open Multipage-PDF (Maybe odd total pages required for bug - not tested)
2. Select Bookview (Ctrl-8)
3. (not checked if necessary) View last two pages.
4. Ctrl-P for printing
5. Select user definded pages
6. Type in the number of the pages the pdf has to print the last page.
7. Print

**Expected behavior**
Last page should be printed.

**File that reproduces the problem**
Multipage-PDF

**Additional context**
In Singlepageview (Ctrl-6) printing the last page works as intended.

### LINKED ISSUE #4817: Printing a single page from the middle of the document prints page 1 of the document
**SumatraPDF version**
3.5.2-64bit

**Describe the bug**
When printing a single page from the middle of a multi-page document, SumatraPDF prints the first page of that document and not the page that I selected.

**To Reproduce**
1. Open a multi-page PDF document
2. Go to 'File > Print' (or press control-P)
3. Click on 'Pages' and select 'Custom Range'
4. Enter a page number from the middle, say 4
5. Click on 'Print'

**Expected behavior**
The 4th page of the PDF file will be printed

**File that reproduces the problem**
Bug is triggered with all files. But I'm still attaching a sample file here.

**Screenshots**
Attaching two screenshots. One, that shows the choices I pick when printing and two, the actual printed page. As is clear from the screenshots, even though I have picked page 4 to print the output is page 1.

**Additional context**
This is Windows 11. SumatraPDF is the only PDF viewer installed.

Image

Image

[SixPager.pdf](https://github.com/user-attachments/files/18978815/SixPager.pdf)

Maxime-J:
Fixes #3453, fixes #4235, fixes #4779, fixes #4817.

Setting a single page print range in modern UI results in SumatraPDF printing the current page.
It's caused by a change of behavior between legacy and modern print dialog, which, like you identified,
sets flags in a non expected way.

Here are flags hex values in different scenarios:

<ins>Single page range (2 or 2-2)</ins>
Legacy 0x00040006
Modern 0x00440014 -> Causing the issue because `PD_CURRENTPAGE` is set.

<ins>Other ranges</ins>
Legacy 0x00040006
Modern 0x00040016

<ins>Current page</ins>
Legacy 0x00440004
Not available in modern

<ins>All pages</ins>
Legacy 0x00040004
Modern 0x00040014

So, modern dialog sets `PD_CURRENTPAGE` instead of `PD_PAGENUMS` when a single page range is entered,
`lpPageRanges` containing the custom range, with `nFromPage` and `nToPage` set to the chosen page.

About code, note that `nPageRanges == 1` check is added for precaution,
it is always equal to 1 when legacy and modern dialogs set `PD_CURRENTPAGE`.

### DIFF
--- a/src/Print.cpp
+++ b/src/Print.cpp
@@ -1,4 +1,4 @@
-/* Copyright 2022 the SumatraPDF project authors (see AUTHORS file).
+/* Copyright 2025 the SumatraPDF project authors (see AUTHORS file).
    License: GPLv3 */
 
 #include "utils/BaseUtil.h"
@@ -901,7 +901,16 @@ void PrintCurrentFile(MainWindow* win, bool waitForCompletion) {
     }
 
     if (pdex.Flags & PD_CURRENTPAGE) {
-        PRINTPAGERANGE pr = {(DWORD)dm->CurrentPageNo(), (DWORD)dm->CurrentPageNo()};
+        PRINTPAGERANGE pr;
+        if (pdex.nPageRanges == 1 && pdex.lpPageRanges[0].nFromPage == pdex.lpPageRanges[0].nToPage) {
+            // Unified Print Dialog (which doesn't have a "Current page" option)
+            // sets PD_CURRENTPAGE when the custom range contains one page (2 or 2-2)
+            // with nFromPage and nToPage equal to the chosen page.
+            // PD_PAGENUMS isn't set in that case.
+            pr = pdex.lpPageRanges[0];
+        } else {
+            pr = {(DWORD)dm->CurrentPageNo(), (DWORD)dm->CurrentPageNo()};
+        }
         ranges.Append(pr);
     } else if (win->CurrentTab()->selectionOnPage && (pdex.Flags & PD_SELECTION)) {
         printSelection = true;


kjk:
thanks!

PR #5117, opened 2025-09-29
Enable multiple instances when using unique -appdata command line option

imero:
When running multiple instances with -appdata pointing to different directory and using "useTabs" setting then only first instance settings are applied.
Example use case is using two shortcuts with different settings files to keep open tabs organised in two instances. If first has for example work related pdf tabs and second has personal tabs open then whichever process is first opened only that restores its open tabs, second shortcut execution only opens new window with 0 tabs not a new process.

After change the -appdata value is always incorporated in hash that is used in instance detection mutex name.

### DIFF
--- a/src/SumatraStartup.cpp
+++ b/src/SumatraStartup.cpp
@@ -402,11 +402,11 @@ static void SetupCrashHandler() {
 }
 
 static HWND FindPrevInstWindow(HANDLE* hMutex) {
-    // create a unique identifier for this executable
+    // create a unique identifier for this executable and appdata combination
     // (allows independent side-by-side installations)
-    TempStr exePath = GetSelfExePathTemp();
-    str::ToLowerInPlace(exePath);
-    u32 hash = MurmurHash2(exePath, str::Len(exePath));
+    TempStr combinedPath = str::Join(GetSelfExePathTemp(), "|", GetAppDataDirTemp());
+    str::ToLowerInPlace(combinedPath);
+    u32 hash = MurmurHash2(combinedPath, str::Len(combinedPath));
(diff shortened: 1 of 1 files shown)


GitHubRulesOK:
the usage of -appdate was only for one session thus you would for 2 concurrent sessions simply use 2 DISTINCT sumatrapdf launch dde.exes this is what I have always proposed with dual forms otherwise its a nightmare with new windows etc. as the DDE channel is likely to be dirty.

one oddity of running with NO history is you can use 2 themes 
image

imero:
Is there any other known parameters or workflow to have more general and cleaner solution for this multi-tab, multi window, tabs restore, instead always remembering to copy-paste-rename or script the executable for each shortcut when building latest code? But I do understand the DDE interop compatibility might be something to consider.

GitHubRulesOK:
Your idea of including -appdata in Mutex is a good one for such dual usage. I am not developer to know how robust that is but would ensure in tests there is no other affected usage for commands such as DDE which may depend ALSO on a distinct identification where the in-Memory driving is via a single Sumatra|Control|Topic thus not Sumatra1 and Sumatra2 (Which would be NICE to Have)

GitHubRulesOK:
@kjk 
Thinking about this the only concern I think is legacy use of DDE master server is  Sumatra and thus that should be retained for a single instance (avoiding other apps such as LaTeX Editors, needing rebuild/recompile.
But if there are 2 or more sessions, then they should be broadcasting as different servers such as Server Sumatra2 Sumatra3 ...

kjk:
makes sense, thanks!

PR #5131, opened 2025-10-11
Modified ToC Navigation and Bookmark searching

Adem-O:
Changed ToC Navigation to use HJKL for bookmark navigation. To avoid conflict with bookmark searching, bookmark searching by letters now uses Shift+<Char> to search. hjkl mapped to arrow keys for vim-esque navigation.

### DIFF
--- a/src/TableOfContents.cpp
+++ b/src/TableOfContents.cpp
@@ -41,12 +41,79 @@
 
 #include "utils/Log.h"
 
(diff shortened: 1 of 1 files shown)


kjk:
Thanks. 

It's an interesting idea but I'm afraid that it changes the behavior of tree view and some people will be confused and will complain.

There doesn't seem to be a way to implement this without changing how search works.

Issue #2202, opened 2021-10-19
Navigating via bookmark requires two Back commands to undo


SpecLad:
1. Open any PDF file with bookmarks ([example](https://webassembly.github.io/spec/core/_download/WebAssembly.pdf)).
2. Click on a bookmark.
3. Press Alt+⬅. Nothing will happen.
4. Press it again. SumatraPDF will navigate back to the beginning of the document.

Expected behavior: SumatraPDF should navigate back in step 3. It shouldn't be necessary to press the shortcut twice.

I can reproduce this with both the current release (3.3.3) and the development branch (commit 68c5a9912ccfc52be1234158391129db06c52e83).

GitHubRulesOK:
that commad is a bit of an oddball in MS trees its a go back to previous thus can jump forwards to previous mark just as much as back to start.

To see that in action pick a few bookmarks in reverse order, the same in windows explorer, the last resting place is the one first opened and in a PDF opened by SumatraPDF that is usually counted as the first page.

SpecLad:
@GitHubRulesOK That's totally fine. The problem is that it doesn't jump _anywhere_ at first.

Anyway, I did a bisect and found out that this is a regression that was introduced in 9f2b9d7d.

SpecLad:
I did some more digging and found the cause of the problem. When you click the bookmark, both `TocTreeClick` and `TocTreeSelectionChanged` run and both of them navigate to the new location. Therefore, two locations get added to the location history instead of one.

`TocTreeClick` should probably only do its thing if the item being clicked is already selected.

humidifier-dev:
@SpecLad Hey! I also investigated this situation and made a pull request. 
I also think that clicking on selected bookmark should only works if this bookmark is already selected. So I've made a pull request
https://github.com/sumatrapdfreader/sumatrapdf/pull/5157

GitHubRulesOK:
@DerDunkelheit 
Current version works as per issue thus was closed how does your addition make that different ?
Use the above sample click on introduction it goes there instantly

<img width="700" height="249" alt="Image" src="https://github.com/user-attachments/assets/619d2126-a134-44cd-a16f-f0214b392951" />

click on back and it goes back to first page. which is correct "Alt + Left Go back"

<img width="700" height="295" alt="Image" src="https://github.com/user-attachments/assets/49ac3829-4dff-44c6-8a4f-202928ac31c2" />

humidifier-dev:
[...]

Issue #1716, opened 2020-07-30
Bookmarks regression


GitHubRulesOK:
Whilst checking issue #1715 
It became apparent that adding a single valid ? bookmark which works in 3.1.2 does not work in recent builds, oddly it works if saved as .bkm or .vbkm !!
I dont have enough versions kept between 3.2.12002 (working) to get any closer but not working since 3.2.12007 (January) pre 3.2 release 

[BlankA4x2 ex VS eXpert.pdf](https://github.com/sumatrapdfreader/sumatrapdf/files/5004024/BlankA4x2.ex.VS.eXpert.pdf)

kjk:
The issue here is that there's only one item in toc tree and technically it's always selected so clicking on it doesn't change the selection and hence we don't act upon it.
This is a result of rewriting toc tree to use TreeCtrl.
Not sure how hard it's to fix. This shouldn't happen often in real PDFs.

GitHubRulesOK:
@kjk 
That definition we dont "Act" on single bookmarks could explain other odd behaviours I experienced on occasions when building vbkm using single page pdfs.
Dont have a repeatable example at present but can keep it in mind as a cause.

PR #5157, opened 2025-10-29
Navigate to selected bookmark

DerDunkelheit:
In this pull request I made FavTreeItemClicked working for currently selected tree item.

When there’s only one bookmark, you can navigate to it only once because navigation to the bookmark page happens in the FavTreeSelectionChanged callback. This callback is triggered only when a different tree element is selected.

In the code, there’s also an onClick handler, but it was commented out in this commit: https://github.com/sumatrapdfreader/sumatrapdf/commit/32b298161e6286d23d8253d874f9f2aca38c65ef

because of two related issues.:
https://github.com/sumatrapdfreader/sumatrapdf/issues/1716 https://github.com/sumatrapdfreader/sumatrapdf/issues/2202

### DIFF
--- a/src/Favorites.cpp
+++ b/src/Favorites.cpp
@@ -775,14 +775,13 @@ void RememberFavTreeExpansionStateForAllWindows() {
     }
 }
 
-#if 0
-static void FavTreeItemClicked(TreeClickEvent* ev) {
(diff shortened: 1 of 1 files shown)


kjk:
Thanks!

PR #5194, opened 2025-12-03
Theme.cpp: add multiple new themes

hnvy:
Hi

Thanks for the great reader.

If you like this, then I've got more planned :)

Harvey


GitHubRulesOK:
There is no need to add to the cpp (adding exe bloat)  those are already allowed for in user choice of add to settings. what might be nice is a HELP page of examples, so user just adds one or two.

image

hnvy:
Ah, my bad! I see that this was implemented to v3.6.

PR #5321, opened 2026-02-12
Update some turkish translations in translations.txt - 

mahirgul:
Update some turkish translations in translations.txt -


GitHubRulesOK:
the normal way to add translations is per https://www.sumatrapdfreader.org/docs/Contribute-translation which will explain how to use https://www.apptranslator.org/app/SumatraPDF

I note there are 22 needing adding @ https://www.apptranslator.org/app/SumatraPDF/tr

kjk:
yeah, translations.txt are auto-generated from https://www.apptranslator.org/app/SumatraPDF so you need to update translations there

Issue #5394, opened 2026-03-20
Just a few typos and omissions in docs


yozhic:
[Here they are: d494765...4000022](https://github.com/yozhic/sumatrapdf/compare/d494765...4000022)
I could submit a pull request, but I'm afraid that's too small of a contribution.

GitHubRulesOK:
I am intrigued about the first one as in my experience SETTINGS use AARRGGBB for highlights for example currently this is mid blue

<img width="533" height="271" alt="Image" src="https://github.com/user-attachments/assets/68ea94a0-0a06-47c4-8a20-97d6f8be64d1" />  

ff as opacity comes first

<img width="544" height="269" alt="Image" src="https://github.com/user-attachments/assets/f9998fcd-92be-4256-a9c6-3d4dd6c002a2" />

yozhic:
Hmm... 🤔  I haven't checked how it works. I was just going by the example that follows later in the text, on [line 570](https://github.com/sumatrapdfreader/sumatrapdf/blob/master/docs/md/Advanced-options-settings.md?plain=1#L570):

> For example #ff0000 means red color. #ff00007f is half-transparent red.

And as we can see, the opacity in this example is the last one...

Well, then we need to fix this example:

> For example `#ff0000` means red color. `#7fff0000` is half-transparent red.

kjk:
Yes, please do submit a PR

yozhic:
[Done.](https://github.com/sumatrapdfreader/sumatrapdf/pull/5397)

kjk:
merged, thanks!

PR #5397, opened 2026-03-21
A few typos and omissions in docs

yozhic:
Continuing the [#5394](https://github.com/sumatrapdfreader/sumatrapdf/issues/5394).
So, in the syntax for color values, the opacity value is specified first (I corrected the example, not the explanation).


kjk:
Thanks!

PR #5467, opened 2026-04-02
Fix MRU tab switching by always adding current tab to the top of the MRU list

dasebasto:
The MRU tab switching is now correct.
Example behavior:
1/ User selects A, then B, C, D
2/ Ctrl-Tab switches from D to C
3/ Ctrl-Tab switches from C to D
4/ Ctrl-Tab,Tab switches from D to B
5/ Ctrl-Tab switches from B to D

Notes:
- MRU tab switching applies to the tabs in the *current* window only (the tabs in the other windows are at the end of the list, in non-MRU order)
- closing a tab does *not* switch to the last tab used (consistent with Firefox MRU behavior)


kjk:
thanks!

Issue #5559, opened 2026-04-20
Click on annotation sholud focus it - even if it rectangle of another annotation


koppor:
**SumatraPDF version**

- Version 3.7.18324

**Describe the bug**

1. Have a text spanning two lines
2. Create an underline annotation covering the first line and some of the second line
3. Create an underline annotation covering the last part of the last line
4. Click on the first annotation
5. Click on the second annotation

Actual behavior: first annotation is selected

**Expected behavior**

Second annotation selected

**Screenshots**

<img width="1758" height="157" alt="Image" src="https://github.com/user-attachments/assets/9c0c409e-66ae-46a4-ad6f-33beb1fb10d6" />

GitHubRulesOK:
HMM the problem in this case is priority of Rectangle Reference locations as a click will find it is inside one or BOTH Rects so the first has natural priority Which will happen with any 2 Annotations if you do the small one before the big one the small one is first candidate 

<img width="368" height="159" alt="Image" src="https://github.com/user-attachments/assets/0ae53b5b-75de-4154-9aab-fa9a8d7c61aa" />

kjk:
Unfortunately not easy to fix. We create bounding rectangle which covers the selected text. For annotations spanning 2 lines we create rect that covers the whole 2 lines. That overlaps with rectangle for second annotation and it just happens to be selected first. Changing that is a big change.

Maybe I can pick overlapping annotations more intelligently.

GitHubRulesOK:
@kjk the annots that "wrap" are a common feature and thus not confined to just highlights  and yes it may be possible to use the MuPDF quads rather than the obj rect but when 2 overlap there is a natural order as set by a user so highlighted fretext should not be blocked by the highlight thus Acrobat in Edge will alow to edit the text.
ALTHOUGH other PDF editors may need the highlight dragged off from the text to allow it to be edited !

[sample.pdf](https://github.com/user-attachments/files/26904620/sample.pdf)

Issue #5558, opened 2026-04-20
"Create Annotation" should focus in "Annotations"


koppor:
**Is your feature request related to a problem? Please describe.**

My aim is to have as less clicks as possible to comment on the PDF. Moreover, I am always frustrated when creating highlight annotations involving a PDF link. Then I need to concentrate when clicking the highlighting (for adding text) and not hitting the link.

**Describe the solution you'd like**

When selecting text and then pressing "A" (for "Create Annotation From Selection"-> "Underline") the annotation should be selected in the "Annotations" dialog - and the focus should be put in the field "Contents".

**Describe alternatives you've considered**

Create Underline Annotation and then click it. However, this 

**Additional context**

When creating a "Text" annotation, the "Annotation" window is focused, the annotation too (which is good), but not the field "Contents" (which would be perfect)

---

Also applies for "Caret" annotations and surely others.

GitHubRulesOK:
The biggest prior users complaint WAS 
It should be Top Level item has focus when opening editor NOT Content as then there is no "common navigation"

Cant be everything and in theory some comments may have no "content" as standalone they don't have to be part of anything certainly they are NOT attached to the graphics on the page and often use a different set of scalars. Whilst this may seem odd annotations are separate REFERENCES to/ from the page but are often bound together and we see the Caret is part of the StrikeOut nothing to do with the page contents. Thus ask for the focus will go to the strikeout.

<img width="346" height="287" alt="Image" src="https://github.com/user-attachments/assets/8d0c4120-a2d2-456c-8586-3b8e2d07163b" /><img width="528" height="257" alt="Image" src="https://github.com/user-attachments/assets/99e7b999-5738-493c-8107-54ae3a5e8f7c" />

kjk:
Different people want different things to be focused. I can't please everyone.

You can do it with `focusedit` option (https://www.sumatrapdfreader.org/docs/Commands#other-cmdcreateannot-arguments) in a custom shortcut definition.

koppor:
Can you help me configuring this?

I have 

```
Shortcuts [
	[
		Cmd = CmdCreateAnnotHighlight openedit
		Key = a
	]
]
```

Nothing happens if annotation window is closed or opened

Also with

    Cmd = CmdCreateAnnotHighlight openedit focusedit

GitHubRulesOK:
@koppor 
[...]

PR #5613, opened 2026-05-07
Focus Contents edit field on annotation click

koppor:
## Summary

- Clicking an annotation (Ctrl+click on the page, or click while the Edit Annotations window is already open) now focuses the `Contents` edit field instead of the annotations list, so a note can be typed immediately.
- The cursor is placed at the end of the existing text rather than selecting all, so additional text can be appended without overwriting.

This addresses the same friction described in #5558 / #5559 - minimizing clicks to add a note to an existing highlight (or any annotation).

## Changes

- `src/Canvas.cpp`: pass `EditAnnotFocus::Edit` when click-to-edit opens or refocuses the annotation in the Edit Annotations window.
- `src/EditAnnotations.cpp`: in `EditAnnotFocus::Edit` mode, position cursor at end of contents instead of selecting all.
- `docs/md/Version-history.md`: changelog entry under `next: 3.7`.

## Test plan

- [ ] Open a PDF with no Edit Annotations window open. Ctrl+click on a highlight -> window opens, `Contents` field is focused, cursor at end of any existing text.
- [ ] With the Edit Annotations window already open, click on a different highlight in the page -> that annotation is selected in the list, focus moves to `Contents`, cursor at end.
- [ ] Selecting an entry in the annotations list (rather than clicking on the page) keeps focus in the list - unchanged behavior.
- [ ] Creating a new annotation via `CmdCreateAnnot*` with `focusedit=true` still focuses `Contents` - unchanged behavior (cursor at end is irrelevant for empty content).

### DIFF
--- a/docs/md/Version-history.md
+++ b/docs/md/Version-history.md
@@ -59,6 +59,7 @@ Available in [pre-release](https://www.sumatrapdfreader.org/prerelease) builds.
 - you can drag&drop images from PDF documents to other applications (web apps, image editors, file explorer etc.)
 - pen/stylus input now works for text selection on Windows tablets
 - fix Edit Annotations window not restoring to the correct monitor in multi-monitor setups
+- focus `Contents` edit field when clicking on an annotation to open or select it in the Edit Annotations window
 - use `GetFileAttributesEx` instead of opening files for change detection on network drives, avoiding Windows Defender re-scans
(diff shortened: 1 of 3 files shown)


koppor:
Think, this is https://github.com/sumatrapdfreader/sumatrapdf/issues/5558#issuecomment-4282475423

PR #5635, opened 2026-05-20
fix file history order number displaying 0 instead of 10

BenK10:
Once 10 files have been opened, the file history menu displays the oldest item as number 0 instead of number 10.
This is caused by some arithmetic involving +1 and modulus.

For this fix to work, the loop bound in function `AppendRecentFilesToMenu` and the modulus operand in function `AddFileMenuItem` must have the same value (they already did, I just made it explicit by replacing a "magic value" with the constant `kFileHistoryMaxRecent`). Otherwise the menu numbers may "wrap around".

ordinal-bug

### DIFF
--- a/src/Menu.cpp
+++ b/src/Menu.cpp
@@ -1225,8 +1225,8 @@ static void AddFileMenuItem(HMENU menuFile, const char* filePath, int index) {
     menuString = ShortenStringUtf8InTheMiddleTemp(menuString, kMaxRunes);
 
     TempStr fileName = MenuToSafeStringTemp(menuString);
-    int menuIdx = (int)((index + 1) % 10);
-    menuString = str::FormatTemp("&%d) %s", menuIdx, fileName);
+    int menuIdx = (int)((index) % kFileHistoryMaxRecent);
+    menuString = str::FormatTemp("&%d) %s", menuIdx + 1, fileName);
     uint menuId = CmdFileHistoryFirst + index;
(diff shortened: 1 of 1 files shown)


GitHubRulesOK:
Hmm I am not the developer but the 0 is AFAIK intentional it is the key on the keyboard after 9

bkusin:
Indeed you can select files with the 0-9 keys when the file history menu is open.
But I didn't know that, and there is no such command on the [shortcuts](https://www.sumatrapdfreader.org/docs/Commands) page.

GitHubRulesOK:
it is probably so old from 20 years ago it never got fully documented and the other old one has traditionally been used without 0
Alt + 1 go to tab 1. Use Alt + 2, etc. up to Alt + 8 Alt + 9 go to last tab

bkusin:
I concluded that, odd as it may look, this is intentional.

If you open the menu with Alt key and navigate that way, the numbers in the file history are underlined, just like other menu shortcuts that appear when menus are accessed this way. In fact, my aesthetic fix breaks the number 0 key shortcut since only the first character in the history section is a shortcut, so the shortcut key for item 10 becomes 1 instead of 0.

I normally use the mouse, so I did not even consider this, nor have I seen any other program do it this way. LibreOffice for example will display number 10 with the 0 underscored as the shortcut key. But it has to because it displays more than 10 recent items.

PR #5719, opened 2026-06-22
Fix RefHover popup region on mupdf per-glyph ink boxes

koppor:
## Problem

The citation/reference hover popup (`RefHover`) cropped to the wrong region on many real PDFs: a clipped narrow strip, or a full-page landscape fallback showing several entries, instead of the single target bibliography entry.

Root cause: the region detectors in `RefHoverDetect.cpp` key off `coords[i].y` as a *line* coordinate, assuming glyphs on one line share a top. `EngineBase::GetTextForPage` returns mupdf's **tight per-glyph ink boxes**, whose tops vary within a line (a period/comma sits well below a capital; ascenders and brackets higher). So the trailing `.` of the previous bibliography entry could fall into the destination band and hijack the entry-start glyph.

(This only reproduces against the real engine — `PyMuPDF`/`fitz` rawdict reports near-uniform tops and hides it.)

## Fix

- **`NormalizeGlyphLines`**: cluster glyphs by *baseline* (`y + dy`, which is stable across a line) and flatten each line to a uniform top-aligned row before detection. `RefHoverOnTimer` normalizes the page coords once and passes them to both `DetectEquationBox` and `DetectEntryBox`.
- **`DetectEntryBox` bracket path** — derive `columnRightX` from the body run, bridging the hanging-indent labelsep gap (capped so it can't jump a column gutter). A narrow biblatex label (`[TA05]`) no longer collapses the box width and clips the entry.
- **`DetectEntryBox` bracket path** — trailing-gap trim so a last-on-page entry (no sibling `[` below it) ends at its last line instead of swallowing the page-number footer.

## Testing

- `RefHover_ut`: added coverage for baseline normalization, the variable-top entry-start hijack, narrow-label full width, 2-column safety, and the footer trim. `test_util` passes all unit tests.
- Verified against the real engine on two biblatex PDFs (single- and multi-page bibliographies): every citation link now yields a correct full-width single-entry popup region.

### DIFF
--- a/src/RefHover.cpp
+++ b/src/RefHover.cpp
@@ -858,11 +858,22 @@ void RefHoverOnTimer(RefHoverState* s, HWND hwndCanvas, EngineBase* engine, floa
         int textLen = 0;
         Rect* coords = nullptr;
         const WCHAR* text = engine->GetTextForPage(destPage, &textLen, &coords);
+        // mupdf's per-glyph ink boxes have varying tops within a line; the
+        // detectors key off coords[i].y as a line coordinate, so flatten each
+        // line to a uniform top-aligned row first (see NormalizeGlyphLines).
+        Rect* normCoords = coords;
+        if (coords && textLen > 0) {
+            normCoords = AllocArray<Rect>((size_t)textLen);
+            NormalizeGlyphLines(coords, normCoords, textLen);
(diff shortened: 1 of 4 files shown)


kjk:
thanks!

PR #5669, opened 2026-06-11
Add citation/reference hover preview (RefHover)

koppor:
Follow-up to https://github.com/sumatrapdfreader/sumatrapdf/pull/5611 with a single commit.

This update is **entirely** written by claude - "only" driven by me in a code-and-fix manager

Screenshot:

grafik

- Closes https://github.com/sumatrapdfreader/sumatrapdf/discussions/3326
- Fixes #128
- Fixes #4221

PR #5720, opened 2026-06-23
Add text-based reference howering

koppor:
Follow-up to https://github.com/sumatrapdfreader/sumatrapdf/pull/5669

Example PDF: https://doi.org/10.1109/WICSA.2015.29

Does not have hyperlinks, I want to hover, too.

This PR adds it.

grafik


kjk:
thanks!

PR #5721, opened 2026-06-23
RefHover: plain wheel scrolls main doc; Shift+wheel scrolls popup

koppor:
Previously plain mouse-wheel over the citation-hover popup zoomed/scrolled the popup, which was surprising while scrolling through a document. Now plain wheel falls through to scroll the main document as if the popup weren't there; Shift+wheel scrolls the popup, Ctrl+wheel zooms it.

See https://github.com/sumatrapdfreader/sumatrapdf/discussions/5671#discussioncomment-17407515


kjk:
thanks!

Issue #1313, opened 2019-12-12
URL Hyperlink bug


motianya211314:
word url no set Hyperlink：
![word](https://user-images.githubusercontent.com/15711438/70719111-5d577280-1d2c-11ea-9b07-5f11b4519def.png)

adobe normal display：
![adobe](https://user-images.githubusercontent.com/15711438/70719183-76f8ba00-1d2c-11ea-8770-27271dca7bff.png)

sumatrapdf :
![sumatrapdf](https://user-images.githubusercontent.com/15711438/70719225-8c6de400-1d2c-11ea-9548-19a56cb3f7f4.png)

GitHubRulesOK:
@motianya211314 
The  sample file (in a zip helps to check structure of the link)

kjk:
@motianya211314  Without a file to inspect I can't tell if it's sumatra or incorrectly generated PDF file.

SumatraPeter:
> here my sample from wordpad shows that it appears to be a problem in the way that MSWord creates such hidden links

Testing with Wordpad is pointless because it tells us _nothing_ about whether Word is or isn't at fault,

I just tried copy-pasting `打开浏览器, 打开 https://www.baidu.com/, 并登` into Word (no hyperlink explicitly added), then used Save As to save as a PDF. Subsequently opened the PDF in both 3.1.2 and prerelease-11625, and neither had a problem with correctly identifying the link to `https://www.baidu.com/`. The comma and characters after the URL were _not_ included in the link by either Sumatra version. In short, I am unable to replicate the issue even with Word.

So @motianya211314 you need to tell us exactly which version of Word you're using, how you converted the document into a PDF, and which version of Sumatra did you test with?

GitHubRulesOK:
@SumatraPeter 
Thanks for cheking in word, although I have (paid) and used in the recent past many clients modern versions  I dont use any version other than my trusty speedy portable 97 copy. 😃

motianya211314:
@SumatraPeter SumatraPDF 3.1.2 x64  Word 2016 and word 2019 save as pdf.
![SumatraPDF](https://user-images.githubusercontent.com/15711438/70764621-09cc3f80-1d93-11ea-90b1-30be9bd47630.png)
[test.docx](https://github.com/sumatrapdfreader/sumatrapdf/files/3958718/test.docx)
[test.pdf](https://github.com/sumatrapdfreader/sumatrapdf/files/3958719/test.pdf)

[...]

PR #5729, opened 2026-06-23
Auto-link plain-text DOIs to https://doi.org

koppor:
In one picture:

grafik

## What

Detects printed DOIs in PDF text (e.g. `10.1109/WICSA.2015.29`) and turns them into clickable `https://doi.org/<doi>` links, alongside the existing auto-linking of plain-text URLs and email addresses.

A DOI is matched structurally: `10.` + a 4–9 digit registrant + `/` + a non-empty suffix. Trailing sentence punctuation (`. , ; : ) ] } '`) is trimmed so a DOI printed inside a sentence or parentheses links cleanly.

## How

Implemented in `LinkifyText` (`src/EngineMupdf.cpp`), the same path that already auto-links `http(s)://`, `www.` and email text. Because of that:

- hover cursor (hand), the URL tooltip and click-to-navigate all work with no extra wiring;
- the `DisableAutoLinks` advanced setting disables DOI auto-linking too (it gates the whole `FzLinkifyPageText` call);
- DOIs already inside an embedded link or a `doi.org/10...` URL are skipped by the existing overlap / preceding-char guards, so no double-linking.

## Notes / limitations

- Single-line only: a DOI wrapped across two lines is not linked.
- `https` is permitted by the default `LinkProtocols`.


kjk (commented):
# Code review: PR #5729 — Auto-link plain-text DOIs

**Verdict: Approve with minor follow-ups** (behavior looks correct; a few maintainability/doc gaps worth addressing before or right after merge)

## Summary

+44 lines in `EngineMupdf.cpp`: adds `LinkifyFindDoiEnd()` and one branch in `LinkifyText()`. Reuses the existing auto-link pipeline (`FzLinkifyPageText`, overlap skip, `DisableAutoLinks`, `https` protocol allowlist). Good fit for the feature.

---

## What works well

- **Right layer.** DOI detection lives next to URL/email linkify helpers, not scattered through canvas/input code.
- **Helper extraction.** `LinkifyFindDoiEnd` mirrors `LinkifyEmailAddress` — testable shape even though it's `static` today.
- **Reuses existing guards.** The `start[-1]` alnum/`/` check prevents matching `…prefix10.…`; overlap skip in `FzLinkifyPageText` avoids doubling embedded links.
- **URL construction is correct.** `protocol = L"https://doi.org/"` + full matched span → `https://doi.org/10.…`, which passes `LinkProtocols` (`https`).

---

## Findings (priority order)

### 1. LinkifyText is another special-case branch on an already busy chain (maintainability)

`LinkifyText` is a growing `else if` ladder (email, http, www, mailto, now DOI). This PR doesn't *create* that problem — `EngineMupdf.cpp` is already ~5k lines — but it does extend it.

**Code-judo option (not blocking for 44 lines, but worth considering):** a small table of `{first char predicate, detectEnd, optional protocol}` would let DOI be one row instead of bespoke control flow. Alternatively, move all `Linkify*` helpers + `LinkifyText` into `Linkify.cpp` so `EngineMupdf.cpp` stops absorbing feature logic.

### 2. Suffix boundary logic diverges from URL linkify (possible CJK edge case)

`LinkifyFindEnd` uses `EndsURL()`, which treats fullwidth comma `U+FF0C` as a terminator (issue #1313). `LinkifyFindDoiEnd` stops on `iswspace` and `"<>"` only — a DOI followed by `，` could swallow the comma into the suffix/rect.

**Suggestion:** reuse `EndsURL(*p)` (or a shared `LinkifyStopsAt(c)` helper) for the suffix scan, then apply DOI-specific trailing-punct trim. Keeps URL/email/DOI behavior consistent without much code.

### 3. Trailing-punctuation trim duplicates `LinkifyFindEnd` logic

Both functions trim closing `.,;:)…` with slightly different rules. Not wrong, but it's the kind of "two places to update" drift that accumulates.

**Suggestion:** extract a tiny `LinkifyTrimTrailingPunct(end, start, charset)` used by both, or call into shared logic after the suffix scan.

### 4. No automated tests for the parser

`LinkifyFindDoiEnd` is pure wide-string parsing — ideal for a small unit test table:

| input | expected end-offset / nullptr |
|-------|-------------------------------|
| `10.1109/WICSA.2015.29` | full match |
| `10.1109/WICSA.2015.29.` | trim trailing `.` |
| `(10.1234/abc)` | trim trailing `)` |
| `10.12/abc` | nullptr (registrant too short) |
| `10.1234567890/abc` | nullptr (registrant too long) |
| `10.1234/` | nullptr (empty suffix) |
| `11.1234/abc` | nullptr |

Even a `tests/ad-hoc-doi-linkify.ts` driving a tiny extracted function (or `-dbg-control` hook) would protect against regressions. Static functions in `EngineMupdf.cpp` make this harder — another reason to consider moving linkify code out.

### 5. User-visible docs not updated

Feature changes auto-link behavior gated by `DisableAutoLinks`, but docs still say "URLs and email addresses" only:

- `docs/md/Hyperlinks.md` (auto-detected links section)
- `docs/md/FAQ.md` / `Advanced-options-settings.md` if they enumerate what gets auto-linked
- next `docs/md/Version-history.md` bullet (project convention for user-visible behavior)

Small doc pass would match the thorough PR description.

### 6. Minor spec nit: `iswdigit` vs ASCII-only registrant

DOI registrant codes are ASCII digits. `iswdigit` accepts other Unicode decimal digits. Very unlikely in extracted PDF text, but `(*p >= '0' && *p <= '9')` would be stricter and cheaper.

---

## Not concerns

- **False positives from `'1'` probe:** `LinkifyFindDoiEnd` bails unless prefix is exactly `10.` — fine.
- **Double-linking `https://doi.org/10.…`:** http branch wins at `'h'`; loop jumps past the span.
- **File crosses 1k lines:** already far past that; this diff doesn't change the calculus, though decomposition is still desirable long-term.
- **Single-line only:** acknowledged in PR; acceptable v1.

---

## Approval bar

No structural regression serious enough to block. I'd merge after (ideally) aligning suffix terminators with `EndsURL` and a short docs update; tests would make this much safer long-term.

kjk:
I've made tweaks as suggested per grok

PR #5730, opened 2026-06-23
RefHover: clickable launch links in popup + better hover lifecycle

koppor:
Makes the citation-hover popup interactive and smooths its show/hide behavior.

Wished by https://github.com/sumatrapdfreader/sumatrapdf/discussions/5671#discussioncomment-17404983

## Changes
- **Clickable launch links**: click an external URL / file link (incl. auto-detected DOI/URL) rendered inside the popup to open it — same path as a direct click in the document (`DocController::HandleLink`). Hand cursor shown over such links.
- **Grace-period hide**: the popup no longer vanishes the instant the cursor leaves the source link, so the cursor can cross the gap onto the popup (e.g. to click a link); it stays while the cursor is over the popup.
- **Reposition on a different occurrence** of the same reference: the skip-render check now also compares the source location. Plain-text citations emit a stable per-occurrence source-line key for this.
- **Instant swap**: when a popup is already showing, hovering a different reference shows the new one immediately instead of re-waiting the full hover delay.

## Files
`src/RefHover.{h,cpp}`, `src/RefHoverText.{h,cpp}`, `src/RefHoverTextDetect.{h,cpp}`, `src/Canvas.cpp`.

Builds clean (dbg64, 0 warnings/0 errors). Unit tests unaffected (new detector param defaults to `nullptr`).


kjk (commented):
# Code review: PR #5730 — RefHover popup interactivity + hover lifecycle

**Verdict: Request changes** — behavior direction is good, but the implementation adds avoidable indirection and pushes `RefHover.cpp` well past a maintainable size boundary.

---

## Summary

Makes the citation popup interactive (click URL/file/DOI links, hand cursor), adds grace-period hide so the cursor can cross the gap onto the popup, compares source location to reposition on duplicate references, and instant-swaps when a popup is already open. Touches the right feature area (`RefHover*`, `Canvas` wiring).

---

## Blockers

### 1. `RefHover.cpp` crosses 1k lines (1000 → 1142)

This file was already at the limit; +143 lines without decomposition is a strong maintainability regression. The PR adds popup hit-testing, hide lifecycle, `hitEngine` retention, and more state on `RefHoverState` — all cohesive, but exactly the kind of chunk that should move out:

- **Extract `RefHoverPopup.cpp`** (wndproc, `LaunchLinkAtPopupPt`, cursor/click handling, `RefHoverScheduleHide` / `RefHoverOnHideTimer`)
- Or extract **link hit-test + page-coordinate mapping** into a small helper used by both cursor and click paths

Please decompose before merge rather than growing the monolith further.

### 2. Click-to-open is a two-hop hit-test with duplicated logic

Flow today:

1. Popup `WM_LBUTTONDOWN` → `LaunchLinkAtPopupPt()` on `hitEngine` → stores `clickPage` / `clickPagePt` → `PostMessage(kRefHoverClickMsg)`
2. Canvas handler → `dm->GetEngine()->GetElementAtPos()` again → `HandleLink()`

The popup already found the launch link; the canvas repeats the same test on a potentially different engine pointer. That's incidental complexity and a boundary leak (popup knows page coords; canvas re-derives the link).

**Code-judo:** add `RefHoverHandlePopupClick(RefHoverState* s, DocController* ctrl, LinkHandler* lh)` (or pass `MainWindow*`) in `RefHover.cpp` that:

- Uses the `IPageDestination*` from `LaunchLinkAtPopupPt` once
- Calls `HandleLink` + `RefHoverHide` there

Canvas keeps a one-liner in the switch, or the popup posts a message that calls that helper. Deletes `clickPage` / `clickPagePt` transient fields and the duplicate kind-check in `Canvas.cpp`.

### 3. `Canvas.cpp` absorbs more RefHover orchestration

`kRefHoverClickMsg` adds a 20-line `HandleLink` block inside the giant canvas switch, plus two `RefHoverScheduleHide` call sites. This is feature logic leaking into an already busy WM handler.

Prefer a single RefHover entry point (`RefHoverOnCanvasMessage` / `RefHoverHandleClick`) so Canvas stays wiring-only.

---

## Significant (non-blocking but worth fixing)

### 4. `srcRect` per-occurrence key is line-level only

```cpp
*srcRectOut = Rect{0, coords[cursorIdx].y, 0, coords[cursorIdx].dy};
```

Two citations of the same reference on **one text line** (e.g. `[1] … [1]`) share the same key → `sameSrc` skip may suppress repositioning. If that's rare in practice, document it; if not, include x/dx (glyph span) or cursor index.

### 5. `sameSrc` manual field compare

`RectF` has `operator==` — `sameSrc` could be `s->displayed.srcPage == srcPage && s->displayed.srcRect == srcRect` (or a small `RefHoverSameSource(...)` helper). Less drift when fields change.

### 6. Coordinate mapping duplicated in `WM_LBUTTONDOWN`

After `LaunchLinkAtPopupPt` succeeds, click handler recomputes `clickPagePt` with the same border/zoom/region math the helper already did. If you keep the canvas round-trip, at least factor `PopupClientToPagePt(s, hwnd, cx, cy)` shared by cursor + click.

### 7. No tests for new behavior

Existing `RefHover_ut` / detector tests unchanged. Worth adding:

- `srcRectOut` populated for numeric + author-year detections
- Unit-level tests for hide/show lifecycle are hard without HWND, but detector `srcRect` keys are easy table tests

---

## What works well

- **Grace hide via `RefHoverOnHideTimer` + `WindowFromPoint` polling** — pragmatic way to keep the popup alive while the cursor transits the gap; cancelling hide on `RefHoverSchedule` is correct.
- **Instant swap (`delayMs = 0` when popup visible)** — good UX, localized change in `RefHoverSchedule`.
- **`hitEngine` AddRef in `RefHoverOnTimer`** — correct lifetime for async render; released in `RefHoverDestroy`.
- **Optional `srcRectOut` on detectors** — backward-compatible default `nullptr`; plain-text path threads it through cleanly.

---

## Minor notes

- Reusing `citationHoverDelay` for hide grace (with 250ms floor) couples two UX knobs; acceptable if documented, but a dedicated hide grace constant might be clearer later.
- `RefHoverScheduleHide` invalidates `renderGen` / pending show — correct; cursor leaving a link should not let a slow render pop the window back up.

---

## Approval bar

Not met yet:

- **File-size explosion** without decomposition (`RefHover.cpp` 1142 lines)
- **Avoidable duplicate hit-test + Canvas switch growth** — clear code-judo path exists
- **Feature orchestration in Canvas** instead of RefHover module

I'd approve after: (1) split or extract popup interaction + hide lifecycle, (2) collapse click handling to a single hit-test in RefHover, (3) thin Canvas wiring.

Issue #5692, opened 2026-06-16
Search: highlight all matches at once, with total count (N of M) and scrollbar position markers


kjk:
Split out from the (now-closed) search umbrella #154. Closely related to the closed #1765 ("Highlight and number matching instances in search results").

**Request:** make in-document search behave like Chrome/Firefox, where *all* occurrences are visible at once, not just the current one. Concretely:

- [x] highlight **all** matches of the current search term simultaneously (the current match emphasized differently from the rest)
- [x] show a **total count and current position**, e.g. `5 / 117`, in the find UI
- [ ] put small **markers on the scrollbar** indicating where matches are in the document
- [ ] (from the original #154 list) allow dismissing an individual highlight (click to de-highlight) and/or keeping highlights persistent

These are grouped because they're one coherent feature (the "search results overview" UX) and would be implemented together. They could complement, or be folded into, the Find All panel (#5691).

r5to:
If it is not a bad idea, i would like to call some glitches and requests here...

1) resizing search window makes refresh glitch on page number:
<img width="265" height="364" alt="Image" src="https://github.com/user-attachments/assets/e8caf8f8-b09b-4e52-b32d-aefee082c42f" />

2) there is lag on search, when results need to scroll search window on the first time scroll. When you make inverse search with `Shift+F3` and again with `F3` - lag is gone. You need to search with `F3` by **holding** it. When you use mouse to scroll up and select first result and try to search with `F3` lag is here again.

3) there is not possible to loop search in search window like without it, in standard simple search (maybe it is not needed (??), but just im telling it works different)

4) would be nice to have old behavior when first result is marked again, like in 3.6 version (sound + highlight):
<img width="242" height="60" alt="Image" src="https://github.com/user-attachments/assets/0a134a24-330f-48fa-b789-f5a5d97ebb9d" />

5) when you have small width window, then search dialog looks like this:

<img width="474" height="72" alt="Image" src="https://github.com/user-attachments/assets/b08043c7-0509-4d9f-9fe8-e6d4652cf4b3" />

kjk:
@r5to thanks for testing, good suggestions.

I've implemented most of them.

For future suggestions, please open new issues to make it easier to track the status of fixing them.

kjk:
Gonna skip scollbar markers and dismissing highlight

PR #5744, opened 2026-06-24
Find: clamp Page Up/Down in results list instead of wrapping (#5742)

### LINKED ISSUE #5742: Search: wrap-around on page up / page down keys
When navigating search results in find floating ui, don't wrap around in a list for page up / page down

nnhhoang:
Fixes #5742.

### Root cause

In the floating search window's results list, `FindWindowWnd::MoveResultSelection`
(`src/FindWindow.cpp`) moves the selection on `Page Up` / `Page Down` by a fixed
`kPage = 10` matches, then on overshoot **wraps around via modulo** (`idx %= n`).

Because the jump is large (not ±1), the wrapped landing point usually falls on the
*wrong side* of where you started. With 13 matches, pressing **Page Down** from match 8
lands on match 5 — i.e. it moves the selection **backward**. `Page Up` is the mirror
image (moves forward). That's the "feels like hitting the random button" the reporter
described:

| start | current `Page Down` | this PR `Page Down` |
|------:|--------------------:|--------------------:|
| 8 / 13 | 5 (jumps **backward**) | 12 (last match) |
| 3 / 13 | 0 (jumps backward) | 12 |

### Fix

Clamp the page jump to the ends instead of wrapping:

- `Page Down` → `min(cur + kPage, n - 1)` (advance a page, stop at the last match)
- `Page Up`   → `max(cur - kPage, 0)` (go back a page, stop at the first match)

Single-step arrow keys (`Up`/`Down`) and `Find Next`/`Prev` keep their intentional
one-at-a-time wrap (last↔first, issue #5692) — only the large page jump is changed.

### Testing

Developed on macOS (no local Windows build). The change is pure index arithmetic, so I
verified it with a standalone test enumerating both the old (wrap) and new (clamp) math
across all start positions: the old code moves the selection the wrong direction in 10 of
13 cases, while the clamped version is monotonic and always in range (`Page Down` never
moves backward, `Page Up` never forward). Build is covered by CI; would appreciate a
maintainer confirming the feel in-app.

### DIFF
--- a/docs/md/Version-history.md
+++ b/docs/md/Version-history.md
@@ -18,6 +18,7 @@ Available in [pre-release](https://www.sumatrapdfreader.org/prerelease) builds.
 - can convert an image to a PDF: right-click an image (or an open image document) and choose `Image / Convert to PDF`, or pick `PDF` in the format drop-down of the Save Image dialog. The new PDF gets `CreationDate`/`ModDate` metadata with the current time and time zone (fixes #949)
 - in the Favorites pane and menu, a favorite for a file with a long name now shows your favorite's name first, then the file name, so the name you gave it is no longer pushed out of view (fixes #829, #2236)
 - case-insensitive search now treats German ß as equivalent to `ss`, so searching `Strasse` finds `Straße` and vice versa (fixes #933)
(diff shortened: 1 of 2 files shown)


kjk:
I don't want untested ai PRs

PR #5769, opened 2026-07-01
RefHover: fix watermarked 2-column pages and wrapped numeric citations

koppor:
Three citation-hover fixes surfaced by a two-column IEEE proof with a diagonal "under review" watermark.

## Watermark strip (2-column popup spanned both columns)
A diagonal watermark crosses the column gutter, so `DetectEntryBox`'s gutter scan never finds an empty strip and `columnRightX` runs into the neighbouring column — the popup box then covers both. `StripWatermarkGlyphs` removes oversized glyphs that sit on sparse, similar-height baselines (a rotated stamp) while keeping headings (dense same-height rows). It runs once on the raw glyphs before `NormalizeGlyphLines`, in `RefHoverShow`, so both detectors see clean text.

## En-dash / em-dash page ranges
`[N]` citation detection accepted only the ASCII hyphen, so a typeset range like `[9–14]` (en-dash) broke bracket detection entirely. Now accepts figure/en/em dash and minus sign too.

## Wrapped numeric citations (`[42,\n43]`)
The index-based walk failed two ways in real PDFs: the glyph stream isn't in reading order across a line break (a wrapped number can sit ~200pt from its `[` in the array), and a naive y-band reconstruction merges both columns (burying the `[` behind the other column's text). `DetectNumericCitationInPageText` now rebuilds local reading order from **column-limited line segments** — the cursor line plus the nearest line above/below in the same column, each restricted to the contiguous x-run split at gutters — so `[42,` connects to `43]` without pulling in the neighbour column or the watermark.

## Tests
Unit tests added for watermark stripping (incl. heading-not-stripped), en-dash ranges, and wrapped citations (end-of-line, out-of-order glyph stream, two-column). Full `test_util` suite green apart from pre-existing unrelated `SettingsUtil`/`StrFormat` failures.

## Not included (parked)
- `[63]`-style entries that flow across the column break (left-column tail → right-column head) — needs a two-crop compose.

### DIFF
--- a/src/RefHoverDetect.cpp
+++ b/src/RefHoverDetect.cpp
@@ -221,6 +221,109 @@ void NormalizeGlyphLines(const Rect* coords, Rect* out, int glyphCount) {
     free(lineId);
 }
 
+// Drop diagonal draft / "under review" watermark glyphs from a page's *raw*
+// glyph arrays before any box detection. A watermark stamp is set far larger
+// than body text and, being rotated, sits roughly one glyph per baseline —
+// each on a sparse row — whereas a heading or title is a horizontal run of
+// same-baseline glyphs. Removing it up front keeps a 2-column gutter empty and
+// the entry bounds tight, instead of special-casing oversized glyphs in every
+// scan. Run this on the engine's raw coords *before* NormalizeGlyphLines:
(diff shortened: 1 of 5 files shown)


kjk:
thanks!

PR #5773, opened 2026-07-02
RefHover: show bracket entries that wrap across a column break

koppor:
## Summary
- Bracket-style bibliography entries ("[63]") that run off the bottom of their column with no sibling "[" or blank-line gap now get detected as wrapping into the top of the next column, and the popup shows both fragments stitched together.
- Fixes a stitching bug found while testing on a real 2-column manuscript PDF: `RenderPage` can return a non-32bpp palette DIB for near-monochrome text crops, so the stitcher now uses GDI `BitBlt` instead of a raw pixel memcpy.
- The continuation search distinguishes a real wrapped tail from page footers/numbers, full-width footer/header lines, and unrelated running body text sharing the page — using column confinement and a tight closing cap rather than a fixed distance-to-page-bottom threshold (real margins vary too much for that to be reliable).
- Wheel zoom/scroll continue to operate on just the primary region as before (untouched code paths) — the stitched continuation only appears on the initial popup show and is dropped on any interaction.
- Click-to-open-link hit test now rejects clicks landing in the stitched continuation strip instead of mis-mapping them to the wrong page point.

## Test plan
- [x] `test_util.exe`: all RefHover unit tests pass (2 new: `ColumnWrapContinuationDetected`, `ColumnWrapContinuationRejectsUnrelatedBodyText`), same 8 pre-existing unrelated failures elsewhere
- [x] Full `SumatraPDF.exe` build succeeds
- [x] Manually verified against a real 2-column IEEE-style manuscript PDF with wrapped bibliography entries (both a synthetic case and two real-world entries, including one requiring several detection fixes for footer/header false positives)

### DIFF
--- a/src/RefHover.h
+++ b/src/RefHover.h
@@ -66,6 +66,15 @@ struct RefHoverState {
         int pageNo = -1;
         float zoom = 0.f;
         RectF region{};
+        // Second crop stitched below `region` in the delivered bitmap, for a
+        // bracket-style entry that wraps across a 2-column page break (see
+        // DetectEntryBox's continuationOut). Empty (dx/dy <= 0) when there's
+        // none. Only ever set by the initial DetectEntryBox-driven show —
+        // wheel-zoom/scroll re-renders build a fresh request from just
+        // displayed.region and never populate this, so the stitched strip is
+        // dropped as soon as the user interacts (region-shift math for a
+        // composited bitmap isn't supported).
+        RectF continuationRegion{};
         // initial show: commit displayed.* and show the popup on completion.
         // false for wheel zoom / scroll re-renders, which update displayed.*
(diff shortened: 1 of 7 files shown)


kjk:
thanks!

koppor:
Thank you for the quick merges! This is kind of code and fix - and I cannot provide test PDFs easily.. - The current one was really strange - links were shown, but not working - because the water-marking destroyed the targets.

koppor:
I think, with this patch, we can do more than any other PDF reader with this regard!

Others might reflow the PDF from two column to single column. But if we are "traditional" and want to keep the two column format (ease of reading, proper layout, ...), then with SumatraPDF, one can really read it well. 💪

PR #5774, opened 2026-07-02
Debug build: log to build dir, not the PDF's folder

koppor:
## Question first
Is this behavior change acceptable? Happy to adjust or drop it — flagging for a maintainer decision rather than assuming.

## Problem
In debug builds the log file defaults to a **relative** `sumlog.txt`. A relative path is resolved against the current working directory, and when a PDF is opened via Explorer the cwd is the **PDF's own folder**. Result: a `sumlog.txt` gets written into every directory the user opens documents from.

## Change
`SumatraStartup.cpp`: in the `gIsDebugBuild` default, leave `flags.logFile` unset instead of forcing `"sumlog.txt"`. Logging then falls through to `GetLogFilePathTemp()` (the build directory, e.g. `out/dbg64/sumatra-log.txt`).

- Debug logging stays **on** by default.
- Log just goes to the build dir instead of littering document folders.
- Passing `-log-to-file <path>` still overrides as before.

## Alternatives considered
- Keep the relative name but make it absolute under `%TEMP%`/appdata.
- Disable debug auto-logging entirely.

Preference?

### DIFF
--- a/src/SumatraStartup.cpp
+++ b/src/SumatraStartup.cpp
@@ -1889,10 +1889,7 @@ int APIENTRY WinMain(_In_ HINSTANCE /*hInstance*/, _In_opt_ HINSTANCE, _In_ LPST
     }
 #endif
 
-    // in debug build, default to logging into the build dir. Use the full path
-    // from GetLogFilePathTemp() rather than a relative "sumlog.txt", which is
-    // resolved against the cwd — when a PDF is opened via Explorer, that is the
-    // PDF's own directory, littering a log file there.
+    // in debug build, default
     if (gIsDebugBuild) {
         if (!flags.logFile) {
             // from the perm arena like all other flag strings (~Flags frees nothing)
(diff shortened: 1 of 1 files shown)


kjk:
better to construct the full path there rather than rely on subtle behavior

koppor:
I thought, there is no real comment seeing now the fix to call an existing method..

koppor:
Claude on testing:

Run: launched with cwd = scratch dir, opened doc.pdf.

  Verified fix:
  - Log written to full path %LOCALAPPDATA%\SumatraPDF-data\f4dd8e\sumatra-log.txt (fresh, 13:03:00, 4 hits on doc.pdf).
  - No sumlog.txt in the PDF's cwd (scratch dir) — no litter. Fix works.

  Note: "build dir" in commit msg = data-dir + exe-sha1 (GetBuildDirNameTemp), not out/dbg64. Behavior correct regardless.

PR #5812, opened 2026-07-19
Fix Linux mingw cross-compile and add a Linux/Wine CI workflow

koppor:
Running on WSL w/ Wine

grafik

---

🤖 *This description was generated by Claude Code.*

## What

`cmd/build-linux-wine.ts` (cross-compile the full SumatraPDF.exe with mingw-w64 on Linux and run it under Wine) had drifted out of sync since it was added: the `mupdf-libs` aggregate removal, the `src/base` rework, and a number of new or renamed `src/` files broke both compile and link. Because no CI exercises this path, nothing caught it. This PR makes the Linux cross-build green again and adds a CI workflow so it stays green.

**Latent portability bugs fixed in source (invisible to MSVC):**
- two case-sensitive `#include` mismatches (`theme.h` in EditAnnotations.cpp, `SumatraPdf.h` in Notifications.cpp) — fatal on any case-sensitive filesystem
- `INT_MAX` used in `Vec.h` without `<limits.h>` (added to `Base.h`; MSVC gets it transitively)
- `Size Size() const;` in `Geom.h` — gcc rejects the unelaborated form ("changes meaning of Size")
- four pointer-truncating `(LONG)hwnd` / `(int)hwnd` casts in `src/uia/` — hard errors under clang; now the canonical `HandleToLong()`
- `SetThreadDescription` isn't declared by mingw-w64 headers before v12 — declared in `WinDynCalls.h` under `#ifdef __MINGW32__` (an identical redeclaration is harmless on newer headers, and the block is invisible to MSVC)
- unrar's internal C++ `Archive` class collides with `src/base/Archive.cpp`'s `Archive` since the base rework — GNU ld errors on the duplicate strong symbols (MSVC silently picks one via COMDAT, a latent hazard there too). unrar TUs now compile with `-DArchive=UnrarArchive`; consumers only use the C API in `dll.hpp`, which doesn't mention the class.

**Shared lib-def drift (`cmd/build-lib-defs.ts` — also benefits the mac/linux native builds):**
- `extract`: add `ext/extract/src` to includes (`memento.h` lives there)
- `libheif`: restore the `heif_components` / `heif_metadata` / `heif_omaf` API files (premake lists them; the TS defs had dropped them → link errors from ExifDump)
- `mupdf`: add `cull-device.c`, `options.c`, `md.c` plus the cmark-gfm include dirs and `CMARK_GFM_STATIC_DEFINE` (previously patched locally in `build-linux.ts` only)

**`cmd/build-with-mingw.ts` re-synced with premake:** the removed `mupdfLibs` import replaced with the individual restored libs; `djvudec`, `a-gumbo`, `zopfli` libs added; ~30 added/renamed src files (AIChat\*, PdfDarkMode\*, PdfCad\*, the RefHover and CommandPalette splits, FindBar/FindWindow, Markdown\*, WebpReader, SumatraLog, AppUnitTests, …); missing include dirs (libjxl, libheif, libwebp, cmark, djvudec); `-Wno-narrowing` for mingw's `QITABENT` macro under clang; the generated TTS/Test stubs re-synced with the current `TextToSpeech.h` / `WStr` signatures; a missing `basename` import.

**Toolchain:** `resolveMingwTools` now prefers `x86_64-w64-mingw32-g++-posix` — Debian's unsuffixed mingw is the win32-threads flavor whose libstdc++ has no `std::mutex` (libheif needs it). On Ubuntu the unsuffixed name is already the posix flavor, so nothing changes there.

**CI — `.github/workflows/linux.yml`:** cross-compiles in a `debian:trixie` container (mingw-w64 ≥ 12 is the practical floor: ubuntu-24.04 runners ship v11, whose headers predate `SetThreadDescription`, `DWMWCP_*`, …), runs the built exe's built-in unit tests under Wine (xvfb) as the gate, runs a render smoke test (`-bench` on a repo PDF, gated on its "Finished" output since `-bench` doesn't return 0 on success), and uploads the exe as an artifact.

## How tested

On Linux, inside a `debian:trixie` container (the same environment the CI job uses):

```
bun cmd/build-linux-wine.ts                                              # -> out/dbg64-wine/SumatraPDF.exe (PE32+)
xvfb-run -a wine out/dbg64-wine/SumatraPDF.exe -for-testing -unit-tests  # "Passed all 24 tests", exit 0
xvfb-run -a wine out/dbg64-wine/SumatraPDF.exe -for-testing -bench tests/issue-3219.pdf
                                                                         # "Finished (in 10.83 ms)"
```

The workflow itself was validated locally with `nektos/act` running the real job in the same container before pushing.

The source changes are written to be no-ops for MSVC (include/decl fixes, `HandleToLong`, a mingw-only `#ifdef` block; the unrar rename lives in the TS lib defs only — premake is untouched), but I could not compile on Windows from this machine — the Windows CI run on this PR is the verification for that.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,0 +1,49 @@
+name: Linux
+on:
+  push:
+  pull_request:
+jobs:
+  wine:
+    name: Cross-compile and test under Wine
+    runs-on: ubuntu-latest
+    # mingw-w64 >= 12 is required (ubuntu-24.04 runners ship v11, whose
+    # headers predate SetThreadDescription, DWMWCP_*, ...)
+    container: debian:trixie
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v7
+
+      - name: Install mingw-w64, Wine and bun
+        run: |
+          apt-get update
+          apt-get install -y --no-install-recommends \
+            g++-mingw-w64-x86-64 wine wine64 xvfb xauth \
+            ca-certificates curl unzip git
+          curl -fsSL https://bun.sh/install | BUN_INSTALL=/usr/local bash
+
+      - name: Cross-compile SumatraPDF.exe
+        run: bun cmd/build-linux-wine.ts
+
+      # WINEPREFIX: the runner sets HOME=/github/home, owned by the runner
+      # uid, but the container runs as root and wine refuses a prefix in a
+      # directory it doesn't own
+      - name: Run unit tests under Wine
+        env:
+          WINEDEBUG: "-all"
+          WINEPREFIX: /tmp/wineprefix
+        run: xvfb-run -a wine out/dbg64-wine/SumatraPDF.exe -for-testing -unit-tests
+
+      # -bench does not return 0 on success, so gate on its output instead
+      - name: Render smoke test under Wine
+        env:
+          WINEDEBUG: "-all"
+          WINEPREFIX: /tmp/wineprefix
+        run: |
+          xvfb-run -a wine out/dbg64-wine/SumatraPDF.exe -for-testing -bench tests/issue-3219.pdf 2>&1 | tee bench.log || true
+          grep -q "Finished" bench.log
+
+      - name: Upload exe
+        uses: actions/upload-artifact@v7
+        with:
+          name: SumatraPDF-linux-wine-debug
+          path: out/dbg64-wine/SumatraPDF.exe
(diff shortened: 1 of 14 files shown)


koppor:
I think, for me, this would be a good start when switching my main OS to Linux 😅.

koppor:
I am not sure about mingw 11 vs. 12, but the "hack" proposed by Claude sounds reasonable to support Ubuntu-24.04, too

kjk:
Thanks. Things have changed so I've made the change based on this PR.

PR #5631, opened 2026-05-18
Add Linux port: GTK 4 GUI with MuPDF rendering engine

iqbaljunaid:
## Summary

This PR adds a native Linux port of SumatraPDF using GTK 4 for the UI and MuPDF for PDF rendering.

## What's included

- **Meson build system** for Linux (compiles all dependencies from source: zlib, brotli, freetype, harfbuzz, libjpeg-turbo, openjpeg, jbig2dec, lcms2, extract, gumbo-parser, libwebp, mujs, libdjvu, MuPDF)
- **GTK 4 GUI application** with:
  - Header bar with open/navigation/zoom controls
  - Keyboard shortcuts (j/k, PgUp/PgDn, Home/End, Ctrl+O, Ctrl+scroll zoom, F11 fullscreen)
  - Drag-and-drop file opening
  - Cairo-based page rendering from MuPDF pixmaps
  - GApplication file handling (open files from command line or file manager)
- **GitHub Actions CI/CD** pipeline (builds on ubuntu-24.04, publishes releases on tags)
- **Packaging** (tar.gz distribution with .desktop file, AppStream metadata, icons)

## Build instructions

```bash
sudo apt-get install -y meson ninja-build gcc g++ libgtk-4-dev
meson setup linux/builddir
meson compile -C linux/builddir
./linux/builddir/linux/src/sumatrapdf
```

## Files added

- `linux/src/main_gtk.c` — GTK 4 GUI application
- `linux/src/meson.build` — Build targets
- `linux/ext/*/meson.build` — Build files for each dependency
- `linux/dist/` — Packaging files (.desktop, metainfo, package.sh)
- `.github/workflows/linux-package.yml` — CI/CD
- `meson.build`, `meson.options` — Root build configuration

No existing Windows code is modified.

PR #5813, opened 2026-07-19
Fix native Linux build and add a native Linux CI job

koppor:
🤖 *This description was generated by Claude Code.*

**Stacked on #5812** (`linux-wine-ci`) — it needs that PR's shared lib-def fixes. GitHub will retarget this to `master` automatically when the base PR merges and its branch is deleted.

## What

`cmd/build-linux.ts` (native Linux build of all dependency libs plus the portable `test_util` / `test_engines` tools) had drifted from premake and from recent source changes:

- `test_util` sources synced with premake: `DocProperties.cpp`, `PdfDarkModeImageRules/Oklab.cpp` and their `_ut` tests — all called unconditionally by `test_util.cpp` and `SumatraUnitTests.cpp`, so the link was broken
- `test_engines`: add the `a-gumbo` dependency lib (`GumboHtmlParser` needs it) and the `PdfCadDetect` / `PdfCadEnhanceDevice` / `PdfDarkModeNoOp` sources premake lists for it
- stale archive names in the link list (`libzlib.a` / `libopenjpeg.a` → `liba-zlib.a` / `liba-openjpeg.a`)
- link `-lcrypto`: `Crypto_posix.cpp` uses OpenSSL's MD5 (the native build now needs `libssl-dev`)
- wrap the `test_engines` archives in `-Wl,--start-group`: harfbuzz is built with custom allocators (`sumatra_hb_*`) whose definitions live in mupdf's `mupdf_load_system_font.c`, behind harfbuzz in link order — single-pass ld can't resolve that without a group

**`ext/a-openjpeg/opj_malloc.h`:** include the x86 intrinsic headers before the `#pragma GCC poison malloc …`. clang — unlike gcc — enforces poison inside system headers, and `mm_malloc.h` (pulled in by the SSE headers that a later chunk of the amalgamation includes) calls `malloc`. Never seen on macOS because arm64 has no `xmmintrin.h`. (Note: regenerating the amalgamation via `cmd/a-openjpeg.ts` would drop this line; its own `validateCompile` step will catch that.)

**CI:** extends `.github/workflows/linux.yml` with a `native` job (same `debian:trixie` container): builds all dependency libs with clang, runs `test_util`'s unit-test suite natively on Linux as part of the build, and links `test_engines`.

## How tested

On Linux, inside a `debian:trixie` container (same as the CI job):

```
bun cmd/build-linux.ts -debug
# -> all dependency libs build
# -> out/linux-dbg64/test_util links and its unit-test suite passes (run by the build)
# -> out/linux-dbg64/test_engines links
```

The workflow job was validated locally with `nektos/act` in the same container. No Windows-facing changes in this PR (`build-linux.ts` is Linux-only; the `opj_malloc.h` change is inside `#if defined(__GNUC__)`).

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -47,3 +47,32 @@ jobs:
         with:
           name: SumatraPDF-linux-wine-debug
           path: out/dbg64-wine/SumatraPDF.exe
+
+  native:
+    name: Native build and unit tests
+    runs-on: ubuntu-latest
+    container: debian:trixie
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v7
+
+      - name: Install clang, OpenSSL headers and bun
+        run: |
+          apt-get update
+          apt-get install -y --no-install-recommends \
+            build-essential clang lld libssl-dev \
+            ca-certificates curl unzip git
+          curl -fsSL https://bun.sh/install | BUN_INSTALL=/usr/local bash
+
+      # builds all dependency libs, runs test_util's unit-test suite
+      # natively, and links test_engines
+      - name: Build dependencies and run native unit tests
+        run: bun cmd/build-linux.ts -debug
+
+      - name: Upload test tools
+        uses: actions/upload-artifact@v7
+        with:
+          name: sumatra-linux-native-tools
+          path: |
+            out/linux-dbg64/test_util
+            out/linux-dbg64/test_engines
--- a/cmd/build-linux.ts
+++ b/cmd/build-linux.ts
@@ -30,6 +30,7 @@ import {
   spawnCmd,
 } from "./build-deps-common";
 import {
+  aGumbo,
   zlib,
   unrar,
   libwebp,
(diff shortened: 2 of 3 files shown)


koppor:
I don't know why this diff is so short when reading https://github.com/sumatrapdfreader/sumatrapdf/pull/5631#issuecomment-4672251158

koppor:
Rendering works - but UI would need to rewritten - we first "wait" for the macOS port - and then work on Linux (I guess)

grafik

kjk:
@koppor I don't really have a preference about doing mac or linux first.

It's just I use mac and not linux, so it's easier for me to work on mac.

If you want to work on Linux UI feel free. You can commit directly, no need to wait for me to review - it's a lot of work and wouldn't be efficient to wait for approval for every change.

Just don't break Windows build.

I think I started on the right foundation for mac/linux port.

My general idea is to replicate the Windows UI but no need to be 100%. Especially in the beginning it's more important to implement features that reproduce windows UI 100%.

I prefer re-using as much code/logic from windows but I understand the realities. It's ok to duplicate, sharing code can always be improved in the future.

PR #5836, opened 2026-07-29
Library and natural speech ereader

nokel:
I've added a Library feature for ebook navigation, and a natural sounding TTS engine to make ebooks sound more like audiobooks if they have TV shows a person can take voices from (or use other voices they want instead)


kjk:
I'm not opposed to those features in general, but:

1. should be separate PRs

2. Chatterbox is not easy to install so it's not a good experience for users

3. Relaying on a python server for the library management is a non-starter. SumatraPDF has to be self-contained

Frankly, I prefer an issue with detailed description of how such feature should work, ui sketches / pointers to how other apps implement this than a PR

nokel:
No problems, I'll try and make a UI sketch and information about how the changes I've made all work and how they should look. The chatterbox integration uses the repo I've made has mostly fixed the trouble I had with the installation of the software by just using powershell to do the heavy work for people without the knowledge, but I understand what you mean, I was a bit lazy with the approach and figured it would be fine because it worked on my computer, obviously the build doesn't work elsewhere so I'll try to get that fixed. That said, I'll start with raising individual issues and break my branches a bit shorter so you don't have to deal with so much at the same time.

PR #5905, opened 2026-08-06
Add EnableReadAloud advanced setting to hide Read Aloud UI

vlakoff:
> This PR was generated with AI assistance (Claude), based on a request in discussion #5670.

Adds an `EnableReadAloud` advanced setting (default `true`) that hides the Read Aloud (text-to-speech) toolbar button, top-level menu, context menu entry, and command palette entries when set to `false`, so people who don't use TTS can keep a leaner UI. Set `EnableReadAloud = false` in `SumatraPDF-settings.txt`, or toggle it from the new **Advanced Settings...** dialog (Ctrl+K), it shows up there automatically since that dialog is built from the settings reflection data.

Suggested here: https://github.com/sumatrapdfreader/sumatrapdf/discussions/5670#discussioncomment-17792459.

Implementation notes:

- `GetCommandVisibility()` in `CommandAvailability.cpp` is the single choke point already used by the toolbar, the main menu, the context menu, and the command palette. Gating the six `CmdReadAloud*` command ids there (behind `!gGlobalPrefs->enableReadAloud`) hides the feature everywhere at once instead of touching four separate code paths.
- The top-level "Read Aloud (TTS)" menu doesn't need an explicit remove check: `BuildReadAloudMenuItems()` just returns early when the setting is off, and `BuildMenuFromDef()` already drops any submenu that ends up empty.
- One subtlety: the context-menu variant of that submenu is tracked through a static `HMENU` (`gReadAloudContextSubmenu`), refreshed on every right-click via `SetReadAloudContextSubmenu()`. Before this change that refresh always happened because the submenu was never empty. With `EnableReadAloud = false` it can now legitimately end up empty, which skips the refresh and would otherwise leave the global pointing at a stale, already-destroyed `HMENU` from a previous right-click. Fixed by resetting it to `nullptr` right before rebuilding the context menu each time.

**Warning:** The `src/Settings.h` file here results from several iterations of AI and manual editing, and the line wrapping doesn't match what a proper generation with `bun cmd/gen-settings.ts` would produce. But I don't want to bloat my computer again by installing Visual Studio just for this.

### DIFF
--- a/cmd/gen-settings.ts
+++ b/cmd/gen-settings.ts
@@ -880,6 +880,12 @@ const globalPrefs: Field[] = [
     .internal()
     .ver("3.7"),
   field(
+    "EnableReadAloud",
+    Bool,
+    true,
+    "if false, hide the Read Aloud (text-to-speech) toolbar button, menu, and command palette entries",
+  ).ver("3.7"),
+  field(
     "ShowToolbar",
     Bool,
     true,
--- a/docs/md/Advanced-options-settings.md
+++ b/docs/md/Advanced-options-settings.md
@@ -118,6 +118,10 @@ ShowTips = true
 ; (e.g. '#ff0000 #00ff00 #0000ff') (introduced in version 3.7)
 CustomColors =
 
+; if false, hide the Read Aloud (text-to-speech) toolbar button, menu, and
+; command palette entries (introduced in version 3.7)
+EnableReadAloud = true
+
 ; legacy bool for toolbar; if Toolbar is empty, derived as show/hide (internal;
 ; use Toolbar instead)
 ShowToolbar = true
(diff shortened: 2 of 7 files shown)


kjk:
I already did something similar: https://github.com/sumatrapdfreader/sumatrapdf/commit/3dc02368b7a48bb4c3a8a5a6f549646124f54630

But I won't hide the menu.

vlakoff:
Ah, I see that commit was pushed only 5 hours ago! No wonder I missed it, especially since I'd actually had this PR sitting around for a while.

My main issue was how prominent the menu item name was, but since you already addressed that in 81078f3b76aaa6736985ac6b5fdfc21a7953426a, I don't really mind the menu entry remaining present. Thanks!

PR #6004, opened 2026-08-20
feat: remove print action from main toolbar

ebrahimng7:
## Summary

- Remove the Print button from the main toolbar.
- Keep the underlying Print command and keyboard shortcut unchanged.
- Preserve the existing toolbar layout and functionality.

## Testing

- Built successfully with Visual Studio 2022.
- Launched the modified application successfully.
- Verified that the Print button is no longer displayed in the main toolbar.
- Verified that the Print functionality remains available.

## Scope

This is a UI-only change. The underlying Print command has not been removed.

### DIFF
--- a/PROJECT.md
+++ b/PROJECT.md
@@ -1,0 +1,26 @@
+# SumatraPDF Enhanced
+
+SumatraPDF Enhanced is a personal open-source project based on
+[SumatraPDF](https://github.com/sumatrapdfreader/sumatrapdf).
+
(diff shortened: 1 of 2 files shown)


kjk:
You can already fully customize the toolbar so no need for that: https://www.sumatrapdfreader.org/docs/Customize-toolbar#choose-which-buttons-are-on-the-toolbar-and-in-what-order

PR #6075, opened 2026-08-27
 show close button on tab hover, not just selected tab

parixiit:


kjk:
that's intentional, as the comment above it says

please don't submit ai-generated PRs

Instead:
* open an issue describing the issue
* if you prototyped this using ai, provide the prompt you used and which ai / model you used

I'll then decide if the functionality is desired and re-run the prompt myself, with necessary modifications

PR #6076, opened 2026-08-27
Add per-tab night mode toggle

parixiit:


kjk:
we're not adding more icons to toolbar and this additional complexity doesn't justify itself: the user can just do Shift + I and add CmdInvertColors button to toolbar using settings, no need to modify the code

also see my other comment about opening issues with description and ai prompt instead of a PR

PR #6139, opened 2026-09-06
Add fast path to CharLowerW()

geekman7473:
At work I open large PDFs quite frequently. The ARM CPU Manual is a good example (~17,000 pages), and it's what I used in my testing. (You can get a copy here: https://documentation-service.arm.com/static/6a43ea3a072af726f8ef38c0?token=)

Searching through PDFs of this size can be somewhat slow. On my machine (Snapdragon X2 Elite) searching through the entire PDF can take ~1,600ms. The is in the "hot" case, where we assume that MuPDF has already extracted the text from all of the pages. That process takes about 20 seconds, and is a different beast entirely.

Having taken ETL traces, the performance bottleneck seems to be calling CharLowerW() many many times while performing "case folding" on the search string and the document text.

In my testing, this change drops the time for a full document search from 1,600ms, to about 500ms.

Here are some before and after graphs from WPA to illustrate the difference:

image

These traces were captured by running a script that searched for 4 different "nonsense" phrases that do not appear in the document, spaced 1 second apart. Note the overall CPU time for the process drops from ~6,600ms to ~2,100ms.

### DIFF
--- a/src/TextSearch.cpp
+++ b/src/TextSearch.cpp
@@ -297,6 +297,20 @@ static int FoldCaseWCharPortable(int c) {
 }
 #endif
 
+#if OS_WIN
+// CharLowerW() wrapper with a fast path for ASCII.
+// In testing, this improved overall search performance by ~3x.
+static int FastCharLowerW(int c) {
+    if (c < 0x80) {
+        if (c >= 'A' && c <= 'Z') {
+            return c + ('a' - 'A');
+        }
+        return c;
+    }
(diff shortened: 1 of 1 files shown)


kjk:
Thanks, I've implemented something similar in https://github.com/sumatrapdfreader/sumatrapdf/commit/fd2a7684d8c0cec674e6ebf1628d74fce8796c5f

geekman7473:
Awesome! Thank you sir.

PR #6147, opened 2026-09-07
Build certain translation units on hot paths with /O2

geekman7473:
Similar to https://github.com/sumatrapdfreader/sumatrapdf/pull/6139, this change also targets search performance when opening large PDFs like the ARM manual. I see in the project history that you are very sensitive to binary size, so rather than building all of Sumatra and MuPDF with /O2, I instead propose selectively applying this optimization to certain TUs that are hot in traces. This improves performance by a couple percent during search, without costing much disk space.

| | Before | After | Delta |
|---|---|---|---|
| Cold first search | 19,270 ms | 18,036 ms | −1,234 ms (−6.4 %) |
| Hot search, best of 5 | 534 ms | 427 ms | −107 ms (−20 %) |
| `SumatraPDF.exe` size | 11,613,696 B | 11,704,832 B | +91,136 B (+0.8 %) |
| `libsumatrapdf.dll` size | 16,522,752 B | 16,617,472 B | +94,720 B (+0.6 %) |

As an aside: I also tried applying /O2 globally and saw that it bloated the full package size by ~2MB which I assumed you would not want.

Also here are WPA traces to show some before and after comparisons. The first screenshot here shows the "cold search" scenario
image

This second screenshot shows 4 successive "hot searches"

image

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -201,6 +201,18 @@ function favor_speed()
   filter {}
 end
 
+-- /O2 for individual translation units that on the hot path.
+-- This is more targeted than favor_speed() which can increase 
+-- binary size more than desired.
+function favor_speed_files(files)
+  for _, f in ipairs(files) do
+    filter { "files:" .. f, "configurations:Release", "platforms:not x64_asan" }
+    optimize "Speed"
+    enablepch "Off"
+  end
+  filter {}
+end
+
 -- per-workspace setting that differ in clang-cl.exe vs cl.exe builds
 function clang_conf()
   filter "options:with-clang"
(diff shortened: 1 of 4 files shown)


kjk:
great work, thanks!

kjk:
I also wonder if instead of doing /O2 on a whole file a more surgical inline/force inline on specific functions that show up hot would work. As I understand it, /O2 just increases inlining.

I also wonder what would PGO do in terms of speed vs. binary size when run on specific scenarios (like search, rendering PDF etc.)

Also wonder if it's possible to use PGO to get a list of hot functions to inline and then apply those inlines in code so that there's no need to do PGO dance on every build.

geekman7473:
RE inlining: I agree that this is more surgical, but I do think it is more brittle than /O2.

PGO is interesting, and I'll admit I don't know how much it impacts binary size. There is a way to extract out of a PGO profile the list of hot functions, I was shown how to do it back when I worked at MSFT, but I have forgotten the recipe. I bet you that an LLM should be able to figure it out though lol. 

To get my list of hot functions I just used WPR to capture a trace, and then had an LLM parse through it to find the source files that corresponded to the hot functions.

kjk:
apparently there's `/d2inlinelogfull` and could do a diff between non-PGO and PGO build.

I recently added `.\cmd\inlined.ts` which extracts inlined functions from .pdb symbols, could also do a diff between PGO and non-PGO.

But it's more complicated because PGO inlines at call site, not just the dumb "inline this function always"

Can be replicated with `[[msvc::forceinline_calls]]` in code but would probably have to build a tool based on disassembly (for every function in binary dump functions it calls) and then compare calee information to deduce where PGO has inlined stuff.

PR #6167, opened 2026-09-11
Proposal: lazy load the Win32 TreeView items in TOC

geekman7473:
When switching tabs between two large PDFs (for example the ARM CPU Manual, and the Intel CPU Manual) there is a noticeable delay. Digging into this in WPA, it looks like the cost is from building and tearing down the table of contents on the left pane. Seemingly, every item in the TOC requires its own Win32k API call to instantiate. Since these PDFs contain thousands of sections, building this TreeView takes a substantial amount of time.

To fix this issue, I propose "Lazy Loading" the Win32 TreeView out of your linked datastructure internal structure. The TreeView has support for this, in that you receive a message of type TVN_ITEMEXPANDING when the user expands an item, giving us an opportunity to fill in the real children on-demand.

I will admit that this implementation was produced with Opus Fable 5.1, and I am not certain that it matches your code style. For that reason, I'm creating this PR as a "draft" assuming that you will want to reimplement something similar yourself.

Here you can see a before and after in WPA, showing the CPU time delta when running a script that switches tabs back and forth rapidly:
image

And here is a video showing the before and after:
https://youtu.be/HdHr2ZNJVZo

### DIFF
--- a/tests/issue-5918.ts
+++ b/tests/issue-5918.ts
@@ -11,10 +11,23 @@
 import { mkdirSync, rmSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 import { cmdId, runStandalone, tmpPath } from "./util";
-import { findChildWindow, getWindowText, postMessage, sendMessage, sleep, WM_CLOSE } from "./winapi";
+import {
+  findChildWindow,
+  getWindowText,
+  postMessage,
+  sendMessage,
+  sleep,
+  treeGetNextItem,
+  TVGN_NEXT,
+  TVGN_ROOT,
+  TVM_EXPAND,
+  WM_CLOSE,
+} from "./winapi";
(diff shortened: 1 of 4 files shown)


geekman7473:
Oh and here's a link to the Intel manual, if you wanted to reproduce my claims: https://cdrdv2.intel.com/v1/dl/getContent/671200

tests/issue-5918.ts
geekman7473, line 1: NB: This test had to change since it inspected the count of the TreeView items directly, which no longer matches expectations due to the lazy loading. To get this passing again, I opted to have the test expand all of the items first, to fully "materialize" the TOC, but there might be a better way to do it.

kjk:
Thanks, landed as https://github.com/sumatrapdfreader/sumatrapdf/commit/0c04378324e6b6ba348091f3700cc02b7d3d5a20

Issue #106, opened 2015-02-20
Color / Black and White / Grayscale


scolland:
Is it possible to add an options to the command line options for color / greyscale?

Just simply color/BW or something similar?

Thanks

gschenck:
+1

ousia:
:+1:

kjk:
Not sure what it means. Rendering documents as black&white? That wouldn't work without dithering.

Rendering documents in gray scale? Not sure what the use of that would be. Are there other apps that are doing something like that?

GitHubRulesOK:
@azumukupoe 
they are usually serving the pdf as low resolution jpg images so not really comparable that would be more line saving in say comicbook format, google docs and FireFox (not Edge) do something similar but more suited to annotation.

GitHubRulesOK:
@kjk potentially done by tools as there are many customisable options but simplest as "Gray"

```
@echo off
cd /d "%~dp0"
REM Place this file in the same folder as standalone SumatraPDF.exe
REM "%~1" is the input file if using another caller like "drag and drop"
REM If provided as %p in external viewers (not via drag and drop) %2 would be a single page (or a range of pages)

%cd%\SumatraPDF-prerel-32.exe recolor -o "%~dpn1-gray.pdf" "%~1" %2
timeout 1 >nul
%cd%\SumatraPDF-prerel-32.exe -new-window "%~dpn1-gray.pdf"
```

kjk:
should be fixed in latest pre-release

PR #6207, opened 2026-09-17
Add grayscale page rendering mode

trovejo:
Adds an optional grayscale rendering mode for fixed-page documents.

Behavior:
- Shift+B toggles grayscale for the current session.
- PDF page contents are rendered in grayscale.
- PDF annotations and widgets keep their original colors.
- Selection, search and Read Aloud overlays remain colored.
- FixedPageUI.Grayscale controls the startup state and defaults to false.
- Shift+B only changes the current session and does not modify the persisted setting.

Implementation:
- For PDF documents, page contents are rendered first and converted to grayscale before annotations/widgets are replayed in color.
- Other fixed-page engines use the RenderCache grayscale fallback.

The default behavior is unchanged unless FixedPageUI.Grayscale is enabled or Shift+B is pressed.

Tested on current master (17e73c536), Release x64 build:
- 0 errors
- grayscale toggle
- colored PDF annotations
- selection/search overlays
- Read Aloud overlays
- interaction with Shift+I
- startup with FixedPageUI.Grayscale = true/false


GitHubRulesOK:
interesting but for reference in part meets/closes  https://github.com/sumatrapdfreader/sumatrapdf/issues/106 but for KJK to consider if a CLI command would be needed?

kjk:
Thanks for the PR. I pushed changes to your branch:

- Merged master and fixed the merge conflict. `CmdToggleGrayscale` is now at the end of the command list, so the ids of existing commands (`CmdNone`, `CmdFileHistory`, `CmdFavorite`) don't shift.
- `CmdToggleGrayscale` now flips and saves `FixedPageUI.Grayscale` instead of being session-only. `UpdateDocumentColors()` copies the setting to the render cache, so editing the settings file also applies it.
- Annotations are now grayscaled too. Grayscale is a single post-render pass in `RenderCache` for all engines, done before the theme recolor. This removed the two-pass PDF rendering in `EngineMupdf`, which skipped the cached display list (slower re-renders) and ran annotations without holding `docLock` (a crash risk if an annotation is edited during a render).
- Removed the Shift+B default shortcut. Users can bind `CmdToggleGrayscale` in `Shortcuts`, and it's available in the Command Palette.
- The command is allowed with no document open, like `CmdInvertColors`.
- Documented it in `Commands.md` and `Version-history.md`.

Kept from the original: `RenderCache::Add()` storing `req.darkModeEpoch`, which fixes a race where a render finishing during a color change was cached as current.

